PR #38, opened 2022-04-01
chore: Make header names that always need to be sent constants

poppoerika:
This is a follow-up pr for the implementation option 2 mentioned [here](https://github.com/momentohq/client-sdk-csharp/pull/37#discussion_r840714313).

### DIFF
--- a/Momento/HeaderInterceptor.cs
+++ b/Momento/HeaderInterceptor.cs
@@ -8,8 +8,9 @@ namespace MomentoSdk
 {
     class Header
     {
-        private const string Agent = "Agent";
-        public readonly List<string>  onceOnlyHeaders = new List<string>{Agent};
+        public const string AuthorizationKey = "Authorization";
+        public const string AgentKey = "Agent";
+        public readonly List<string>  onceOnlyHeaders = new List<string>{Header.AgentKey};
         public string Name;
         public string Value;
         public Header(String name, String value)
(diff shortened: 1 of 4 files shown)


Momento/HeaderInterceptor.cs
```diff
@@ -2,11 +2,14 @@
 using Grpc.Core;
 using Grpc.Core.Interceptors;
 using System.Collections.Generic;
+using System.Linq;
 
 namespace MomentoSdk
 {
     class Header
     {
+        private const string Agent = "Agent";
```
gautamomento, line 11: I dont think we should add this in each file. Define once and reference in each class.

e.g. drop the constant private here
```
const string AGENT_KEY = "Agent";
const string AUTHORIZATION_KEY = "Authorization";
```

And then in DataGrpcManager reference in `Headers.AGENT_KEY`

gautamomento: And just one note, you probably should looking up naming conventions for C# for constants `AGENT_KEY` is what is done in Java. So, check the appropriate C# guideline.

poppoerika: It seems like `FullProductName` is its naming convention for constants.
https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/const

gautamomento (approved):
Thank you for doing this.

LGTM

PR #42, opened 2022-04-04
feat: pre-signed URLs for GET/SET

virratanasangpunth:


Momento/MomentoSigner.cs
```diff
@@ -0,0 +1,79 @@
...
+        /// <returns></returns>
+        public string CreatePresignedUrlForGet(string hostname, string jwkJsonString, string cacheName, string cacheKey, uint expiryEpochSeconds)
+        {
+            var payload = CommonJwtBody(cacheName, cacheKey, expiryEpochSeconds);
+            payload.Add("method", new string[] { "get" });
+
+            var encodedJwt = CreateJwtToken(jwkJsonString, payload);
+
+            return $"https://{hostname}/cache/get/{cacheName}/{cacheKey}?token={encodedJwt}";
+        }
```
kvcache, line 28: I think this should be at least 2 methods, `SignAccessToken()` and `CreatePresignedUrlForGet()`.

The sign method should return a JWS that can be used in our sdk to interact with Momento. I realize atm we bake the endpoint into the URL. We need to take it out, and when we do this access token will also work in the grpc direct client.

The CreatePresignedUrlForGet/Set method can call the SignAccessToken method and then simply do the string format on the result.

Momento/MomentoSigner.cs
```diff
@@ -0,0 +1,79 @@
...
+        /// <summary>
+        /// Create a pre-signed HTTPS URL to retrieve an object.
+        /// </summary>
+        /// <param name="hostname">Hostname of the SimpleCacheService. Use the value returned from CreateSigningKey's response.</param>
+        /// <param name="jwkJsonString">JWK in JSON string. Use the value returned from CreateSigningKey's response.</param>
+        /// <param name="cacheName">The name of the Cache.</param>
+        /// <param name="cacheKey">The key of the Object.</param>
+        /// <param name="expiryEpochSeconds">The timestamp that the pre-signed URL is valid until.</param>
+        /// <returns></returns>
+        public string CreatePresignedUrlForGet(string hostname, string jwkJsonString, string cacheName, string cacheKey, uint expiryEpochSeconds)
```
gautamomento, line 20: Looking at the accessibility modifiers for this method. It seems like customers will be calling this method directly. How are they going to provide the hostname? The hostname is tied into our Momento Auth token and customers have no knowledge yet about how to parse that out.

gautamomento: I am also quite curious as to why this isn't a method in `SimpleCacheClient`?

virratanasangpunth: re: why this isn't a method in SimpleCacheClient

https://gomomento.slack.com/archives/C01AC7JT370/p1649105256165299?thread_ts=1649103168.349359&cid=C01AC7JT370

virratanasangpunth: > The hostname is tied into our Momento Auth token

This is an intentional design choice for the pre-signed url implementation. We are going to return the hostname as a part of the CreateSigningKeyResponse: https://github.com/momentohq/client-sdk-rust/blob/main/src/response/create_signing_key_response.rs#L8

gautamomento: So after creating a signing key, customers will need to tie (via code or configuration) signing key and the hostname in someway. This allows us to decouple the pre-signed url endpoints from the cache service. Just that we are putting onus on customers now. 

I see pros and cons to both approaches. We can start with this and learn from feedback 👍

Momento/MomentoSigner.cs
```diff
@@ -0,0 +1,79 @@
...
+
+        /// <summary>
+        /// Create a pre-signed HTTPS URL to upload an object.
+        /// </summary>
+        /// <param name="hostname">Hostname of the SimpleCacheService. Use the value returned from CreateSigningKey's response.</param>
+        /// <param name="jwkJsonString">JWK in JSON string. Use the value returned from CreateSigningKey's response.</param>
+        /// <param name="cacheName">The name of the Cache.</param>
+        /// <param name="cacheKey">The key of the Object.</param>
+        /// <param name="expiryEpochSeconds">The timestamp that the pre-signed URL is valid until.</param>
+        /// <param name="ttlSeconds">The timestamp that the object is valid until.</param>
```
wandaitzuchen, line 38: We should probably have this as the duration in seconds that the item will be stored in cache so that our grpc requests are consistent. Also, it needs to be parsed to duration in milli and given as a query param https://github.com/momentohq/client_protos/blob/main/proto/httpcache.proto#L21
I see that on the design doc there's ttl, cache and key in the cliam. I don't think we need them actually.  Mr2 is getting them from the grpc request already (mapped from name and key in the path, and the query param ttl)

danielamiao: so I think cache, key, and TTL are included in the claim for authorization purposes. Any client can try to access a different cache, key, or change their own TTL in the request/query param, but if it's included in the claim MR2 can verify (via the signature) that the URL was generated for that particular resource and the client isn't trying to access things outside its permitted scope.

wandaitzuchen: Makes sense! Thanks!

virratanasangpunth: +1 to Daniela said about authorization.

Also I'm going to keep this as `ttlSeconds` to be consistent with the existing `SimpleCacheClient` in this SDK.

wandaitzuchen: Yup I understand the purpose of having these fields in the claim now. 
Regarding `ttlSeconds`, I was referring to the doc. It's not `The timestamp that the object is valid until`, but the ttl duration.  To be consistent we can use `Time to Live for the item in Cache.`?
We should still parse `ttlSeconds` to milli twhen adding ttl to payload, so that it matches our proto definition. Just like the existing SDK behavior for grpc get requests https://github.com/momentohq/client-sdk-csharp/blob/main/Momento/ScsDataClient.cs#L166.

virratanasangpunth: Oh right! Good catch on both of those.

virratanasangpunth: Fixed.

Momento/MomentoSigner.cs
virratanasangpunth: Done.

Momento/MomentoSigner.cs
```diff
@@ -0,0 +1,102 @@
...
+using System.IdentityModel.Tokens.Jwt;
+using Microsoft.IdentityModel.Tokens;
+using MomentoSdk.Exceptions;
+
+namespace MomentoSdk
+{
+    public class MomentoSigner
+    {
+
+        private readonly JwtHeader jwtHeader;
```
kvcache, line 11: nit: When using the greatest of all brace styles as you are here (Allman is love), there is absolutely no reason to have leading empty newlines after your Allman { line.

Momento/MomentoSigner.cs
```diff
@@ -0,0 +1,102 @@
...
+        /// <param name="signingRequest">The parameters used for generating a pre-signed URL</param>
+        /// <returns></returns>
+        public string SignAccessToken(SigningRequest signingRequest)
+        {
+            var payload = CommonJwtBody(signingRequest.CacheName, signingRequest.CacheKey, signingRequest.ExpiryEpochSeconds);
+            switch(signingRequest.CacheOperation)
+            {
+                case CacheOperation.GET:
+                {
+                        payload.Add("method", new string[] { "get" });
```
kvcache, line 60: Incorrect indentation (here and elsewhere)

Momento/MomentoSigner.cs
```diff
@@ -0,0 +1,102 @@
...
+            }
+
+            return CreateJwtToken(payload);
+        }
+
+        private string CreateJwtToken(JwtPayload jwtPayload)
+        {
+            try
+            {
+                var jwtToken = new JwtSecurityToken(this.jwtHeader, jwtPayload);
```
kvcache, line 82: Incorrect indentation the other way

kvcache (commented):
formatting nits. Might want a linter soon.

Momento/MomentoSigner.cs
```diff
@@ -0,0 +1,102 @@
...
+            };
+        }
+
+        /// <summary>
+        /// Create a pre-signed HTTPS URL.
+        /// </summary>
+        /// <param name="hostname">Hostname of the SimpleCacheService. Use the value returned from CreateSigningKey's response.</param>
+        /// <param name="signingRequest">The parameters used for generating a pre-signed URL</param>
+        /// <returns></returns>
+        public string SignAccessToken(SigningRequest signingRequest)
```
wandaitzuchen, line 53: Minor: documentation doesn't match

Momento/MomentoSigner.cs
```diff
@@ -0,0 +1,102 @@
...
+        public string CreatePresignedUrl(string hostname, SigningRequest signingRequest)
+        {
+            var jwtToken = SignAccessToken(signingRequest);
+            var cacheName = signingRequest.CacheName;
+            var cacheKey = signingRequest.CacheKey;
+
+            return signingRequest.CacheOperation switch
+            {
+                CacheOperation.GET => $"https://{hostname}/cache/get/{cacheName}/{cacheKey}?token={jwtToken}",
+                CacheOperation.SET => $"https://{hostname}/cache/set/{cacheName}/{cacheKey}?ttl={signingRequest.TtlSeconds * 1000}&token={jwtToken}",
```
wandaitzuchen: `ttl=` -> `ttl_milliseconds=`

virratanasangpunth: ah yes good catch!

Momento/MomentoSigner.cs
virratanasangpunth: > formatting nits. Might want a linter soon.

Looks like the canonical way to do this is to bake the style-guide into the vs project and do `Edit -> Format -> Format Document`

kvcache (approved):
lgtm

PR #44, opened 2022-04-13
fix: url-encode cache name and key in presigned urls

wandaitzuchen:
Closes https://github.com/momentohq/client-sdk-csharp/issues/43

### DIFF
--- a/MomentoTest/MomentoSignerTest.cs
+++ b/MomentoTest/MomentoSignerTest.cs
@@ -72,6 +72,20 @@ namespace MomentoTest
         }
 
         [Fact]
+        public void TestUrlEncoding()
+        {
+            MomentoSigner signer = new MomentoSigner(RS256_JWK);
+            uint expiryEpochSeconds = uint.MaxValue;
+            var url = signer.CreatePresignedUrl("foobar.com", new SigningRequest("testCacheName", "///", CacheOperation.GET, expiryEpochSeconds));
+
+            Uri uriResult;
+            bool result = Uri.TryCreate(url, UriKind.Absolute, out uriResult);
+            Assert.True(result);
+            Assert.Equal(Uri.UriSchemeHttps, uriResult.Scheme);
+            Assert.StartsWith("https://foobar.com/cache/get/testCacheName/%2f%2f%2f?token=", url);
+        }
+
+        [Fact]
         public void TestJwkError()
         {
             uint expiryEpochSeconds = uint.MaxValue;
--- a/MomentoTest/MomentoSignerTest.cs
+++ b/MomentoTest/MomentoSignerTest.cs
@@ -76,13 +76,14 @@ namespace MomentoTest
         {
             MomentoSigner signer = new MomentoSigner(RS256_JWK);
             uint expiryEpochSeconds = uint.MaxValue;
-            var url = signer.CreatePresignedUrl("foobar.com", new SigningRequest("testCacheName", "///", CacheOperation.GET, expiryEpochSeconds));
+            var testCacheKey = "#$&\\'+,/:;=?@[]";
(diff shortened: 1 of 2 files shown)


MomentoTest/MomentoSignerTest.cs
```diff
@@ -71,6 +71,20 @@ public void TestPresignedUrlForSet()
...
+        {
+            MomentoSigner signer = new MomentoSigner(RS256_JWK);
+            uint expiryEpochSeconds = uint.MaxValue;
+            var url = signer.CreatePresignedUrl("foobar.com", new SigningRequest("testCacheName", "///", CacheOperation.GET, expiryEpochSeconds));
+
+            Uri uriResult;
+            bool result = Uri.TryCreate(url, UriKind.Absolute, out uriResult);
+            Assert.True(result);
+            Assert.Equal(Uri.UriSchemeHttps, uriResult.Scheme);
+            Assert.StartsWith("https://foobar.com/cache/get/testCacheName/%2f%2f%2f?token=", url);
```
virratanasangpunth, line 85: Can we do a more exhaustive test? I know some library won't percent encode `/` char.

e.g. https://github.com/momentohq/client-sdk-python/commit/262ac0fa655eff80f781fc135f578bf81cf487ba#diff-5f9e7ab8de4ab1f67260ff9b6ad5a4b3a46f4e469c2f7a6f30fee41d2fc8bb40R96

wandaitzuchen: We don't need to encode `!`, `*`, `(` and `)` (.NET actually won't convert them https://docs.microsoft.com/en-us/dotnet/api/system.net.webutility.urlencode?view=net-6.0#remarks)

Tested with a token for `"key": "._-*!()"`, it worked without url-encoding them. 
```
curl https://cache.developer-wanda-dev.preprod.a.momentohq.com/cache/set/cache/._-\*\!\(\)\?token\=eyJhbGciOiJSUzI1NiIsImtpZCI6Imx5anp3ZGF2bSJ9.eyJjYWNoZSI6ImNhY2hlIiwibWV0aG9kIjpbIkdFVCIsIlNFVCJdLCJleHAiOjI2NDk4MDU2MjUsInR0bCI6MzYwMCwia2V5IjoiLl8tKiEoKSJ9.JCrQ6GtrRrQSzNioF5EU_xcY9v1xpOPMKl1JnS5UAgSbAyU6wAFGE8r0UwxIxNDNEKPRmgy84w97gZavJAf3nCJpHOPP9GAKm5ipQhK-ZCZbt6Ok9IvRnX_Uc23-aiDH10av0QwybQ8l-GTbCkE7U-biWihzD5Yv35zBzkpUOf-fgCzSUIR3PyspcrTE483dDowc_nYbSXJjfZzwdhZa1Yk8hRy4TAWMjh8LHqW2wZ-e8lQpnRbdt7NVtD7KjPQaL0GNNd4yjO33bcak2LqDCTJDQtjtfwk_DkpcU46SCw9fh9nhhRzW6BuWn0BaqUt8oatWHGLCOrAMCqwLsyO3mA\&ttl_milliseconds\=3600000 -d '"'`echo something | base64`'"'
{
 "result": "Ok",
 "message": ""
}
```

wandaitzuchen: Btw in the python SDK example (e.g. https://github.com/momentohq/client-sdk-python/blob/main/tests/test_momento_signer.py#L86) I think we need to escape `\`?  
`%5C` (i.e. encoded `\`) is missing here https://github.com/momentohq/client-sdk-python/blob/main/tests/test_momento_signer.py#L92

virratanasangpunth: Oh ha good catch, I need to double-escape that backslash.
image

virratanasangpunth (approved):

PR #49, opened 2022-04-21
fix: add both `netstandard2.1` and `netstandard1.2`

### LINKED ISSUE #48: Make it compatible with older .NET Frameworks (specifically `netstandard1.2`)
Our SDK packages target `netstandard2.1`. Some history, .NET runtime ships in 2 flavors:
1. .NET Framework (4.8 and lower) which ships with the OS, and runs only on Windows
2. .NET Core, which runs on cross-plat and typically what new apps are written in. .NET Core was rebranded to use the .NET name so after .NET Core 3.1 came .NET 5. 

**Problem:**
.NET Standard 2.1 is only supported on .NET Core, which means customers who are on .NET Framework cannot use our SDK.

**Recommended solution:**
- Target `netstandard1.2`, we have a customers needing support for older versions of .NET Framework (v4.5.2), according to https://docs.microsoft.com/en-us/dotnet/standard/net-standard?tabs=net-standard-2-0 we need to target .NET Standard 1.2 for that.

poppoerika:
Closes #48

### DIFF
--- a/Momento/MomentoSdk.csproj
+++ b/Momento/MomentoSdk.csproj
@@ -1,6 +1,7 @@
 <Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
+    <TargetFramework>netstandard1.2</TargetFramework>
     <TargetFramework>netstandard2.1</TargetFramework>
   </PropertyGroup>
 


danielamiao:
@poppoerika we might have to get you a windows VM instance to test this first. The grpc library dependency we use does not support older versions of .NET so I suspect this change alone will not work properly.

Momento/MomentoSdk.csproj
```diff
@@ -1,6 +1,7 @@
 <Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
+    <TargetFramework>netstandard1.2</TargetFramework>
```
gautamomento, line 4: Not sure how this comes into play. My most immediate concern is how can we test this with specific version. Similar to how we do for python - we test with different python versions.

gautamomento, line 4: Not sure how this comes into play. My most immediate concern is how can we test this with specific version. Similar to how we do for python - we test with different python versions.

danielamiao (changes requested):
Specifically, I think we at least need a Windows VM with an older version of Windows like Windows 8 so we can test with the older .NET framework. Wyze is on .NET framework 4.5.2 which I think is available on Windows 8

gautamomento (commented):
Left a comment around testing. My concern is gRPC version and compatibility with older versions. May be some way we can take this and run on the environment with older .net only?

poppoerika:
@danielamiao @gautamomento 
I found this blog post saying that support for .NET Framework 4.5.2, 4.6, 4.6.1 will end on April 26th (next Tuesday).
https://devblogs.microsoft.com/dotnet/net-framework-4-5-2-4-6-4-6-1-will-reach-end-of-support-on-april-26-2022/
Is Wyze going to keep using .NET Framework 4.5.2 after the support ends?

danielamiao:
@poppoerika They might keep running it for a while, definitely won't be upgraded by next Tuesday.

poppoerika:
@danielamiao @gautamomento 
I wanted to share what I have found so far regarding making our SDK availabe for older .NET frameworks.

Before updating this SDK, we need to update https://github.com/momentohq/client_protos/tree/main/csharp to also include `netstandard1.2`. 
Also we need to remove https://github.com/momentohq/client_protos/blob/main/csharp/Momento.csproj#L26 so that our proto generated files will be compatible with `netstandard1.2`. 
As you mentioned above `Grpc.Net.Client` is only compatible with `.Net 5.0` or higher and ` .NET Standard 2.0` or higher.
And workaround for this is to use `Grpc.Core` https://github.com/grpc/grpc-dotnet/issues/625
I'm not sure how much work this would be but both updating our csharp proto and SDK only using `Grpc.Core` on a Windows EC2 instance would take some time to implement and test. (this is also I'm not familiar with .NET development).

cc/ @eaddingtonwhite

danielamiao:
@poppoerika Thanks for the writeup, your plan makes sense to me. I think it's important to build for both .net standard 2.1 and .net standard 1.2. The latter is older, more outdated, but necessary for some customers. We should probably have a modern version (2.1) like we do now in addition.

poppoerika:
@danielamiao @gautamomento @eaddingtonwhite 
I found this blog from gRPC regarding the plan for `Grpc.Core` we're trying to use instead of `Grpc.Net.Client`.
Especially [here](https://grpc.io/blog/grpc-csharp-future/#the-plan), it's planed to be deprecated next month (May 2022).
I understand that being compatible with the older .Net frameworks is important for some customers but just wanted to bring this up since this means that other customers who use newer frameworks/.Net/.Net Core are affected in a way as well since my understanding is that being compatible with older/newer frameworks/.Net/.Net Core means that using `Grpc.Core` instead of `Grpc.Net.Client` both in our proto generated files and SDK.

danielamiao:
@poppoerika Maybe we should just maintain a fork of our SDK where it supports `Grpc.Core` and `.netstandard1.2`. We can prioritize forking this and releasing it (manual github action) onto nuget to block current customers. This would leave `main` as-is.

poppoerika:
@danielamiao - Yeah, that sounds good to me. I think we'll need to make a fork of the client_proto repo as well to be compatible with the forked C# SDK. 
I'll work on the proto repo first and move on to the SDK repo.

poppoerika:
I'm going to close this pr since we need to fork this repo and open a pr against that fork.

PR #53, opened 2022-05-06
feat: Add multi get

### LINKED ISSUE #52: Add `multi_get`
Reference: https://github.com/momentohq/client-sdk-python/commit/c65565ac1164fe8f20994754ce9508adedf4bbbe

poppoerika:
Add `MultiGetAsyc` following the example of `multi_get` in the Python SDK.

Closes #52

### DIFF
--- a/Momento/Responses/CacheMultiGetFailureResponse.cs
+++ b/Momento/Responses/CacheMultiGetFailureResponse.cs
@@ -1,0 +1,16 @@
+﻿using System;
+
+namespace MomentoSdk.Responses
+{
+    public class CacheMultiGetFailureResponse
+    {
+        public byte[] Key { get; private set; }
+        public Exception Failure { get; private set; }
+
+        public CacheMultiGetFailureResponse(byte[] key, Exception faiure)
+        {
+            Key = key;
+            Failure = faiure;
+        }
+    }
+}
--- a/Momento/ScsDataClient.cs
+++ b/Momento/ScsDataClient.cs
@@ -6,7 +6,7 @@ using CacheClient;
 using Google.Protobuf;
 using Grpc.Core;
 using System.Collections.Generic;
-using System.Text;
+using System.Linq;
 
 namespace MomentoSdk
 {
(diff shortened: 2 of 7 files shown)


Momento/Responses/CacheMultiGetFailureResponse.cs
```diff
@@ -0,0 +1,16 @@
+﻿using System;
+
+namespace MomentoSdk.Responses
+{
+    public class CacheMultiGetFailureResponse
+    {
+        public byte[] Key { get; private set; }
+        public Exception Failure { get; private set; }
+
+        public CacheMultiGetFailureResponse(byte[] key, Exception faiure)
```
malandis, line 10: Second argument has a typo: `faiure` should be `failure`
```suggestion
        public CacheMultiGetFailureResponse(byte[] key, Exception failure)
```

Momento/Responses/CacheMultiGetFailureResponse.cs
```diff
@@ -0,0 +1,16 @@
...
+        {
+            Key = key;
+            Failure = faiure;
```
malandis, line 13:
```suggestion
            Failure = failure;
```

Momento/ScsDataClient.cs
```diff
@@ -69,11 +71,86 @@ public async Task<CacheSetResponse> SetAsync(string cacheName, string key, byte[
...
+        {
+            List<CacheGetResponse> successResponses = new();
+            List<CacheMultiGetFailureResponse> failedResponses = new();
+            foreach (string key in keys)
+            {
+                _GetRequest request = new _GetRequest() { CacheKey = Convert(key) };
+                DateTime deadline = DateTime.UtcNow.AddSeconds(this.dataClientOperationTimeoutSeconds);
+                try
+                {
+                    _GetResponse resp = await this.grpcManager.Client().GetAsync(request, new Metadata { { "cache", cacheName } }, deadline: deadline);
```
eaddingtonwhite, line 89: With how we are currently looping through each key and then awaiting an async call here this will end up being a pretty slow call since its doing them 1 at a time. I think we will want to use async tasks and the WhenAll method for allowing us to fire off all the requests in parallel before awaiting their responses.

https://docs.microsoft.com/en-us/dotnet/api/system.threading.tasks.task.whenall?view=netframework-4.7.1

eaddingtonwhite (approved):
Nice this is looking good to me now! Maybe have @wandaitzuchen take pass as well get another pair 👀 on it as well.

eaddingtonwhite:
@poppoerika is test failure because of issue in alpha still?

poppoerika:
@eaddingtonwhite - It failed because of the alpha issue, but it looks like the issue has been resolved?
At least, now it's passing all the tests.

Momento/ScsDataClient.cs
```diff
@@ -5,33 +5,34 @@
...
 
 namespace MomentoSdk
 {
     internal sealed class ScsDataClient : IDisposable
     {
         private readonly DataGrpcManager grpcManager;
         private readonly uint defaultTtlSeconds;
-        private readonly uint dataClientOperationTimeoutSeconds;
-        private const uint DEFAULT_DEADLINE_SECONDS = 5;
+        private readonly uint dataClientOperationTimeoutMilliseconds;
```
wandaitzuchen, line 16: just curious why switching to milli?

Momento/ScsDataClient.cs
```diff
@@ -59,24 +60,102 @@ public async Task<CacheSetResponse> SetAsync(string cacheName, string key, strin
...
+                {
+                    successResponses.Add(t.Result.SuccessfulResponse());
+                }
+                if (t.Result.FailedResponse() is not null)
+                {
+                    failedResponses.Add(t.Result.FailedResponse());
+                }
+            }
+
+            return new CacheMultiGetResponse(successResponses, failedResponses);
```
wandaitzuchen, line 101: nit: feel like this is a bit much code repetition. would be nice to clean this up a bit. just a nit.

wandaitzuchen (approved):

Momento/ScsDataClient.cs
poppoerika: We thought that more control over request timeout might be needed since we're offering a performant cache product. Also matching with the Python SDK as well as making the failure test possible to make sure multi get accept failed responses as argument.
cc @eaddingtonwhite

wandaitzuchen (approved):
🚢

PR #56, opened 2022-05-19
feat: Add delete method.

### LINKED ISSUE #50: [New API] Add Delete API
Our backend server now supports delete, this functionality needs to be added into the SDK.

Need to update to use newest proto: https://github.com/momentohq/client_protos/blob/main/proto/cacheclient.proto#L35

malandis:
Closes #50 

Update the gRPC client and add support for the data client's delete method. Example usage is as follows:
`client.Delete(cacheName, key)
`

And the corresponding async implementation:
`await client.Delete(cacheName, key)`


tylerburdsall (approved):
🚢

eaddingtonwhite (approved):
🚢

PR #57, opened 2022-05-19
feat: Add dictionary stubs

malandis:
Addresses #55.

Initial dictionary stubs for `DictionarySet`, `DictionaryGet`, and `DictionaryGetAll`. To address with customer and refine per feedback.

### DIFF
--- a/Momento/Incubating/SimpleCacheClient.cs
+++ b/Momento/Incubating/SimpleCacheClient.cs
@@ -21,7 +21,7 @@ namespace MomentoSdk.Incubating
             throw new NotImplementedException();
         }
 
-        public Task<CacheDictionarySetResponse> DictionarySetAsync(string cacheName, string dictionaryName, Dictionary<string, string> dictionary)
+        public async Task<CacheDictionarySetResponse> DictionarySetAsync(string cacheName, string dictionaryName, Dictionary<string, string> dictionary)
         {
             throw new NotImplementedException();
         }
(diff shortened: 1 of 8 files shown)


Momento/Incubating/SimpleCacheClient.cs
```diff
@@ -0,0 +1,51 @@
...
+        public SimpleCacheClient(string authToken, uint defaultTtlSeconds) : base(authToken, defaultTtlSeconds)
+        {
+        }
+
+        public SimpleCacheClient(string authToken, uint defaultTtlSeconds, uint dataClientOperationTimeoutSeconds) : base(authToken, defaultTtlSeconds, dataClientOperationTimeoutSeconds)
+        {
+        }
+
+        // TODO: Dictionary can be <string|byte[], string|byte[]>
+        public CacheDictionarySetResponse DictionarySet(string cacheName, string dictionaryName, Dictionary<string, string> dictionary)
```
eaddingtonwhite, line 19: @malandis Could we also do `DictionarySet(string cacheName, string dictionaryName, Dictionary<string, byte[]> dictionary)` method. i think having string only for keys is ok to start but allowing byte values seems valuable. WDUT

Momento/Incubating/SimpleCacheClient.cs
```diff
@@ -0,0 +1,51 @@
...
+        {
+            throw new NotImplementedException();
+        }
+
+        public async Task<CacheDictionarySetResponse> DictionarySetAsync(string cacheName, string dictionaryName, Dictionary<string, string> dictionary)
```
eaddingtonwhite, line 24: should we add method docs here as well?

Momento/Incubating/SimpleCacheClient.cs
eaddingtonwhite: chatted on zoom more remembering previous convo now. Going to keep string string to start and get some customer feed back.

Momento/Incubating/SimpleCacheClient.cs
eaddingtonwhite: doing in follow up pr

eaddingtonwhite (approved):

PR #62, opened 2022-06-17
fix: Cache client should be in the MomentoSdk namespace.

malandis:
This commit changes the `SimpleCacheClient` namespace from `MomentoSdk.Responses` to `MomentoSdk`.


bruuuuuuuce (approved):
:shipit:

PR #73, opened 2022-07-02
chore: Rename MultiGet and MultiSet to GetMulti and SetMulti.

malandis:
The previous naming convention preferred modifiers before commands, as
in `MultiGet`, `DictionaryMultiGet`. This ordering makes it
difficult to group commands under the IDE and in documentation. The
new ordering `GetMulti`, `DictionaryGetMulti` eases this.

With this way, now the following will be completions of
`DictionaryGet`:

`DictionaryGet`, `DictionaryGetMulti`, `DictionaryGetAll`

### DIFF
--- a/Momento/Incubating/README.md
+++ b/Momento/Incubating/README.md
@@ -24,7 +24,7 @@ class Driver
             key: "my-key", value: "my-value", ttlSeconds: 60, refreshTtl: false);
 
         // Set multiple values
-        scc.DictionaryMultiSet(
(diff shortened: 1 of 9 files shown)


eaddingtonwhite (approved):

cprice404 (approved):
left a few comments offline.  i think that in the near future these methods may get renamed again, and we are likely to pull this incubating stuff out to a separate package.  but as long as everyone is aware of those forthcoming changes this looks fine

PR #79, opened 2022-07-20
chore: enable nullable reference types in projects

### LINKED ISSUE #80: type safety: enable nullable reference types

malandis:
Closes #80 

Enable nullable reference types (NRT) and null static analysis in the
projects. With this setting the compiler issues warnings when
variables/arguments are in the wrong null context. This PR enables the setting,
resolves remaining NRT warnings, and removes previously per-file
nullable-enabled directives.

NRT enabled is best practice in new projects and documented here:

https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/compiler-messages/nullable-warnings
https://docs.microsoft.com/en-us/ef/core/miscellaneous/nullable-reference-types
https://docs.microsoft.com/en-us/dotnet/csharp/nullable-migration-strategies

### DIFF
--- a/Momento/Incubating/Responses/CacheDictionaryGetAllResponse.cs
+++ b/Momento/Incubating/Responses/CacheDictionaryGetAllResponse.cs
@@ -1,5 +1,4 @@
-﻿#nullable enable
-using System;
+﻿using System;
 using System.Collections.Generic;
 using System.Text;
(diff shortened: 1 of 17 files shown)


tylerburdsall (approved):
Nice, this is super neat!

PR #81, opened 2022-07-20
chore: use client test fixture to speedup tests

malandis:
This PR introduces a client test fixture. The previous tests
instantiated the client in test class constructors. That is, we used
this pattern:

```csharp
public class MyTests
{
	public MyTests()
	{
		// Instantiate the client
	}

	public void Test_Something()
	{
		// tests using the client
	}

	public void Test_Otherthing()
	{
		// tests using the client
	}
}
```

Data instantiated in the constructor will be setup and torn down
for *each* test method, not just for the class as a whole. Because
an isolated client is not necessary for most of the tests and incurs
high overhead, we should avoid this.

In `Fixtures.cs` we create a client fixture and register it to be
available across all test classes. We updated the test cases accordingly.

See [the official documentation](https://xunit.net/docs/shared-context) for reference

### DIFF
--- a/MomentoIntegrationTest/Fixtures.cs
+++ b/MomentoIntegrationTest/Fixtures.cs
@@ -1,0 +1,46 @@
+namespace MomentoIntegrationTest
+{
+    /// <summary>
+    /// A cache client fixture.
+    /// Use this when not testing client-building edge cases:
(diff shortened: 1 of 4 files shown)


tylerburdsall (approved):
Nice, looks a lot better!

PR #82, opened 2022-07-20
chore: use guids instead of ad-hoc strings in tests

malandis:
This PR replaces ad-hoc test data with GUIDs. The code is easier to
maintain and less error prone (keys leaking from one test to the
other).

We also refactored test functions to a separate `Utils.cs` file.


tylerburdsall (approved):
GUIDs are always a good idea 💪

PR #91, opened 2022-07-29
chore: prefer ienumerable over dictionary + other style issues

malandis:
This changes the client API to use the broadest interfaces where possible. `IEnumerable<KeyValuePair>` affords the end-user more options than `IDictionary` without limiting the implementation.

While updating this, we corrected the implementation of `CacheSetMultiResponse` to not propagate user-data. And while doing this made some minor style changes to the data client.


tylerburdsall (approved):

kvcache (approved):
I like this. Make sure you get another human to agree though, the dictionary change is subjective =)

kvcache:
I see Tyler is +1 too so shipit

PR #93, opened 2022-08-02
feat: implement set data structure

### LINKED ISSUE #90: set API in C# SDK

malandis:
Closes #90 

Implement the set data structure, including: `SetAddAsync`, `SetFetchAsync`, and `SetDeleteAsync`.

### DIFF
--- a/Momento/Incubating/Internal/ScsDataClient.cs
+++ b/Momento/Incubating/Internal/ScsDataClient.cs
@@ -513,6 +513,37 @@ internal sealed class ScsDataClient : ScsDataClientBase
         return new CacheSetAddResponse();
     }
 
+    public async Task<CacheSetAddBatchResponse> SetAddBatchAsync(string cacheName, string setName, IEnumerable<byte[]> elements, bool refreshTtl, uint? ttlSeconds = null)
+    {
+        return await SendSetAddBatchAsync(cacheName, setName, elements.Select(element => element.ToByteString()), refreshTtl, ttlSeconds);
+    }
+
+    public async Task<CacheSetAddBatchResponse> SetAddBatchAsync(string cacheName, string setName, IEnumerable<string> elements, bool refreshTtl, uint? ttlSeconds = null)
+    {
+        return await SendSetAddBatchAsync(cacheName, setName, elements.Select(element => element.ToByteString()), refreshTtl, ttlSeconds);
+    }
+
+    public async Task<CacheSetAddBatchResponse> SendSetAddBatchAsync(string cacheName, string setName, IEnumerable<ByteString> elements, bool refreshTtl, uint? ttlSeconds = null)
+    {
+        _SetUnionRequest request = new()
+        {
+            SetName = setName.ToByteString(),
+            RefreshTtl = refreshTtl,
+            TtlMilliseconds = ttlSecondsToMilliseconds(ttlSeconds)
+        };
+        request.Elements.Add(elements);
+
+        try
+        {
+            await this.grpcManager.Client.SetUnionAsync(request, MetadataWithCache(cacheName), deadline: CalculateDeadline());
+        }
+        catch (Exception e)
+        {
+            throw CacheExceptionMapper.Convert(e);
+        }
+        return new CacheSetAddBatchResponse();
+    }
+
     public async Task<CacheSetFetchResponse> SetFetchAsync(string cacheName, string setName)
     {
         _SetFetchRequest request = new() { SetName = setName.ToByteString() };
--- a/Momento/Incubating/Internal/ScsDataClient.cs
+++ b/Momento/Incubating/Internal/ScsDataClient.cs
@@ -528,4 +528,23 @@ internal sealed class ScsDataClient : ScsDataClientBase
         }
         return new CacheSetFetchResponse(response);
     }
+
+    public async Task<CacheSetDeleteResponse> SetDeleteAsync(string cacheName, string setName)
+    {
+        _SetDifferenceRequest request = new()
+        {
+            SetName = setName.ToByteString(),
+            Subtrahend = new() { Identity = new() }
+        };
+
+        try
+        {
+            await this.grpcManager.Client.SetDifferenceAsync(request, MetadataWithCache(cacheName), deadline: CalculateDeadline());
+        }
+        catch (Exception e)
+        {
+            throw CacheExceptionMapper.Convert(e);
+        }
+        return new CacheSetDeleteResponse();
+    }
 }
--- a/Momento/Incubating/Internal/ScsDataClient.cs
+++ b/Momento/Incubating/Internal/ScsDataClient.cs
@@ -28,7 +28,6 @@ internal sealed class ScsDataClient : ScsDataClientBase
         return SendDictionarySet(cacheName, dictionaryName, field.ToByteString(), value.ToByteString(), refreshTtl, ttlSeconds);
     }
 
-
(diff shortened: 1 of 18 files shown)


tylerburdsall (approved):

tylerburdsall (approved):

Momento/Incubating/Internal/ScsDataClient.cs
```diff
@@ -482,4 +481,101 @@ public async Task<CacheDictionaryRemoveFieldsResponse> DictionaryRemoveFieldsAsy
...
+        return await SendSetAddAsync(cacheName, setName, element.ToByteString(), refreshTtl, ttlSeconds);
+    }
+
+    public async Task<CacheSetAddResponse> SetAddAsync(string cacheName, string setName, string element, bool refreshTtl, uint? ttlSeconds = null)
+    {
+        return await SendSetAddAsync(cacheName, setName, element.ToByteString(), refreshTtl, ttlSeconds);
+    }
+
+    private async Task<CacheSetAddResponse> SendSetAddAsync(string cacheName, string setName, ByteString element, bool refreshTtl, uint? ttlSeconds = null)
+    {
```
kvcache, line 496: why doesn't this just delegate to SetAddBatchAsync?

Momento/Incubating/Responses/CacheSetFetchResponse.cs
```diff
@@ -0,0 +1,46 @@
...
+    public CacheSetFetchResponse(_SetFetchResponse response)
+    {
+        Status = (response.SetCase == _SetFetchResponse.SetOneofCase.Found) ? CacheGetStatus.HIT : CacheGetStatus.MISS;
+        elements = (Status == CacheGetStatus.HIT) ? response.Found.Elements : null;
+    }
+
+    public HashSet<byte[]>? ByteArraySet
+    {
+        get
+        {
```
kvcache, line 25: this is pretty expensive. Don't you want to cache it in case of repeated calls?

Why is HashSet better than Set for this property's exposed type?

Momento/Incubating/Responses/CacheSetFetchResponse.cs
```diff
@@ -0,0 +1,46 @@
...
+            }
+            return new HashSet<byte[]>(
+                elements.Select(element => element.ToByteArray()),
+                Utils.ByteArrayComparer
+            );
+        }
+    }
+
+    public HashSet<string>? StringSet()
+    {
```
kvcache, line 38: Why is this a method while ByteArraySet is a property?

Momento/Responses/CacheGetResponse.cs
```diff
@@ -1,27 +1,24 @@
...
     {
-        if (Bytes == null)
-        {
-            return null;
-        }
-        return Encoding.UTF8.GetString(Bytes);
+        get => cacheBody != null ? cacheBody.ToByteArray() : null;
     }
+
+    public string? String() => (cacheBody != null) ? cacheBody.ToStringUtf8() : null;
```
kvcache, line 23: Why is this a method while Bytes is a property?

kvcache (commented):
pretty good, got a few curiosities

Momento/Incubating/Internal/ScsDataClient.cs
malandis: I didn't want to wrap the single element in a list as a performance optimization. If you think that's premature I can change it.

Momento/Incubating/Responses/CacheSetFetchResponse.cs
malandis: I'll add in the caching. Didn't have an intuition as to the # of calls we would expect vs extra references in the class.

HashSet: in C# we have `ISet`, the set interface, and concrete types `HashSet`, `SortedSet`, and `Immutable{Hash,Sorted}Set`. In the code the pattern I've followed is: methods should accept arguments with the broadest interface possible (prefer `elements` as `IEnumerable<T>` to `ISet<T>`), but when we return a data structure, the concrete type is known, so specify that.

Momento/Incubating/Responses/CacheSetFetchResponse.cs
malandis: (Same question came up on the dictionary PR): because string conversion could raise an exception. I'm going to make a spec in the near future where we can agree on this in particular.

Momento/Responses/CacheGetResponse.cs
malandis: https://github.com/momentohq/client-sdk-csharp/pull/93#discussion_r938262036

Momento/Incubating/Internal/ScsDataClient.cs
kvcache: seems likely that `new []{ element }` won't be the slow part of this but of course you could always take the builder as a parameter to the internal method instead and completely elide the copy. I don't think it matters much. Less code is better though!

Momento/Incubating/Responses/CacheSetFetchResponse.cs
kvcache: 2 calls seems like  a reasonable thing for a user to do:
```
if (2 < response.ByteArraySet.Count) {
  Console.WriteLine("a set of 3 or more: {0}", response.ByteArraySet);
}
```

and probably it's reasonable to call it K times in the general case. 1 reference in memory is worth reducing n * K to n.

Cool - maybe you should return ImmutableHashSet then?

malandis: Added caching for `ByteArraySet` and `StringSet` in [004d5fb](https://github.com/momentohq/client-sdk-csharp/pull/93/commits/004d5fb5a70afbc1e0a2f5113dcf6069fe31bd68)

Momento/Incubating/Internal/ScsDataClient.cs
malandis: Fixed in [9187dfb](https://github.com/momentohq/client-sdk-csharp/pull/93/commits/9187dfbfb29fcd518a1cf9de5cd1ffa6ac3c2e89)

Momento/Incubating/Responses/CacheSetFetchResponse.cs
```diff
@@ -27,10 +29,15 @@ public CacheSetFetchResponse(_SetFetchResponse response)
...
-            return new HashSet<byte[]>(
+            if (_byteArraySet != null)
+            {
+                return _byteArraySet;
+            }
+            _byteArraySet = new HashSet<byte[]>(
                 elements.Select(element => element.ToByteArray()),
                 Utils.ByteArrayComparer
             );
+            return _byteArraySet;
```
kvcache, line 40: this works but are you aware of https://docs.microsoft.com/en-us/dotnet/framework/performance/lazy-initialization#implementing-a-lazy-initialized-property

also, if you do this as a guard it would be cleaner:

```suggestion
            if (_byteArraySet == null)
            {
                _byteArraySet = new HashSet<byte[]>(
                    elements.Select(element => element.ToByteArray()),
                    Utils.ByteArrayComparer
                );
            }
            return _byteArraySet;
```

malandis: Awesome. Thanks for the lazy initialization pointer and reminder about using a guard. I committed lazy initializers to [59b036e](https://github.com/momentohq/client-sdk-csharp/pull/93/commits/59b036edf624045c1e3cf574556f29d33a3ab096)

kvcache (approved):
looks good thanks

PR #100, opened 2022-08-08
chore: remove synchronous API

malandis:
Remove the synchronous `Get`, `Set`, and `Dictionary` methods. The end user can invoke the async APIs using `await` for by calling `.Result` on the Task object. Because the synchronous gRPC calls perform sync-over-async, there is no safety benefit to exposing synchronous methods in our API.


tylerburdsall (approved):

kvcache (approved):
cool, I think this is better

PR #102, opened 2022-08-09
chore: update documentation

malandis:
This enables documentation in the SDK assembly. By default the `dotnet` tooling does not enable this. We have enabled this and also completed documentation for the cache client and signer classes.

In a future PR we will complete documentation for all public classes and methods in the SDK assembly.


bruuuuuuuce (approved):
:shipit:

PR #106, opened 2022-08-11
chore: temporarily disable CS1591 / missing documentation

malandis:
In a future PR we will address public-facing documentation. For now we
will suppress the multitudinous warnings by disabling the warning
project wide.


eaddingtonwhite (approved):
@malandis looks fine for now. Can we just make sure we have a follow up ticket made in backlog to come back and fix so we don't loose track.

PR #108, opened 2022-08-11
chore: update protos dependency

malandis:
This updates to the latest protos. These properly namespace the protos
to `Momento.Protos`, hence the updates necessary to the client code
here.


eaddingtonwhite (approved):
🧼 like the new namespaces on protos. :shipit:

PR #115, opened 2022-08-22
feat: list push, pop, and fetch

### LINKED ISSUE #110: list apis in C# SDK
List APIs:

- List push front/back
- List pop front/back
- List fetch

malandis:
Add `ListPushFrontAsync`, `ListPushBackAsync`, `ListPopFrontAsync`, `ListPopBackAsync`, `ListFetchAsync` to incubating client.

This PR adds the above methods, tests, and documentation to the client.

We also refactored various unary get response classes to inherit from a common base. In the case of simple `Get`, `DictionaryGet`, and `ListPop`, the state and behavior of unary get responses are the same; the only differences lay in building the response object from the various gRPC response messages.

Closes #110


anp13 (approved):
LGTM.

PR #128, opened 2022-08-29
Move source code to src/Momento.Sdk.

malandis:


eaddingtonwhite (approved):
Looks good to me. A lot of changes but seems right 👍 I like logically how it's broken out more now. Seems like a more standard/sane layout as well under src folder.

PR #130, opened 2022-08-30
docs: use documentation tags where appropriate

### LINKED ISSUE #127: use C# documentation tags

malandis:
Use the breadth of .NET documentation tags where appropriate. The previous code had markdown in places which will not display correctly in Intellisense.

This also reduces a lot of copy-paste documentation by using `inheritdoc`.

Closes #127


bruuuuuuuce (approved):
:shipit: love it!

eaddingtonwhite (approved):
Wow interesting did not know you could do this w/ c#. TIL c# suprises me in new ways every day. :) This looks cleaner/easier to maintain to me! Nice work! we should make sure to double check in IDE as well make sure all looks good to end user w/ this new approach.

malandis:
I verified in VS Code but we should followup in Visual Studio and Rider as well.

PR #134, opened 2022-09-01
feat: implement `ListLength`

### LINKED ISSUE #133: list length client implementation

malandis:
Implements, tests, and documents `ListLengthAsync`. A list not in the
cache is interpreted to have length 0.

Closes #133

### DIFF
--- a/src/Momento.Sdk/Incubating/SimpleCacheClient.cs
+++ b/src/Momento.Sdk/Incubating/SimpleCacheClient.cs
@@ -658,6 +658,23 @@ public class SimpleCacheClient : ISimpleCacheClient
         return await this.dataClient.ListRemoveAllAsync(cacheName, listName, value);
     }
 
+    /// <summary>
+    /// Calculate the length of a list in the cache.
+    ///
+    /// A list that does not exist is interpreted to have length 0.
+    /// </summary>
+    /// <param name="cacheName">Name of the cache to perform the lookup in.</param>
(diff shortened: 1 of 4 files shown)


src/Momento.Sdk/Incubating/SimpleCacheClient.cs
```diff
@@ -658,6 +658,23 @@ public async Task<CacheListRemoveAllResponse> ListRemoveAllAsync(string cacheNam
         return await this.dataClient.ListRemoveAllAsync(cacheName, listName, value);
     }
 
+    /// <summary>
+    /// Calculate the length of a list in the cache.
+    ///
+    /// A list that does not exist is interpreted to have length 0.
```
anp13, line 664: Should we return the list is missing response if it does not exist? What do we do for ListFetch if the list does not exist?

malandis: A list of length 1 will be in the cache. If you pop a value, it will have length zero which also means it is removed from the cache. So length zero for a missing list is consistent with this.

ListFetch has a `Status` property (`HIT`/`MISS`) alongside the payload

bruuuuuuuce (approved):
:shipit:

src/Momento.Sdk/Incubating/SimpleCacheClient.cs
malandis: To be continued!

PR #136, opened 2022-09-02
chore: rename `ListRemoveAll` to `ListRemoveValue`

malandis:
Renaming for clarity: `ListRemoveValue` removes all instances of `value` from a list.

### DIFF
--- a/src/Momento.Sdk/Incubating/SimpleCacheClient.cs
+++ b/src/Momento.Sdk/Incubating/SimpleCacheClient.cs
@@ -639,23 +639,23 @@ public class SimpleCacheClient : ISimpleCacheClient
     /// <param name="value">The value to completely remove from the list.</param>
     /// <returns>Task representing the result of the cache operation.</returns>
     /// <exception cref="ArgumentNullException">Any of <paramref name="cacheName"/>, <paramref name="listName"/>, or <paramref name="value"/> is <see langword="null"/>.</exception>
-    public async Task<CacheListRemoveAllResponse> ListRemoveAllAsync(string cacheName, string listName, byte[] value)
+    public async Task<CacheListRemoveAllResponse> ListRemoveValueAsync(string cacheName, string listName, byte[] value)
     {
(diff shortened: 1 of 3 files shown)


src/Momento.Sdk/Incubating/SimpleCacheClient.cs
```diff
@@ -639,23 +639,23 @@ public async Task<CacheListFetchResponse> ListFetchAsync(string cacheName, strin
...
         Utils.ArgumentNotNull(value, nameof(value));
 
-        return await this.dataClient.ListRemoveAllAsync(cacheName, listName, value);
+        return await this.dataClient.ListRemoveValueAsync(cacheName, listName, value);
     }
 
-    /// <inheritdoc cref="ListRemoveAllAsync(string, string, byte[])"/>
-    public async Task<CacheListRemoveAllResponse> ListRemoveAllAsync(string cacheName, string listName, string value)
+    /// <inheritdoc cref="ListRemoveValueAsync(string, string, byte[])"/>
+    public async Task<CacheListRemoveAllResponse> ListRemoveValueAsync(string cacheName, string listName, string value)
```
eaddingtonwhite, line 652: just thinking out loud but I feel like inevitably someone is going to want a `RemoveValueOnce` vs `RemoveValueAll` ik just doing a rename but the new name even more so makes me think there should be a way to remove just one element in the list of a certain type.

eaddingtonwhite: The way this is named now I would sort of expect this to just remove 1 element instead of all and then maybe for us to add a `filter` function to list as well?

malandis: We will provide the options `ListRemoveValueFront(value, count)` to mean "remove the first `count` occurrences" and `ListRemoveValueBack(value, count)` to mean "remove the last `count` occurrences.

eaddingtonwhite (approved):

PR #147, opened 2022-09-13
feat: release from release branch

cprice404:
Updates the `cd.yaml` job so that we don't do a release on
every commit.  Introduces a new manual release job that we
can use to decide when to do a release.  Releases will come
from the `release` branch rather than the `main` branch.

### DIFF
--- a/.github/workflows/on-push-to-release-branch.yaml
+++ b/.github/workflows/on-push-to-release-branch.yaml
@@ -1,0 +1,63 @@
+name: On push Main
+
+on:
+  push:
+    branches: [release]
+
+jobs:
+  build_csharp:
+    runs-on: ubuntu-latest
+    env:
+      TEST_AUTH_TOKEN: ${{ secrets.ALPHA_TEST_AUTH_TOKEN }}
+      TEST_CACHE_NAME: client-sdk-csharp
+
+    steps:
+      - name: Get current time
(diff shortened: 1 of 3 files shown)


eaddingtonwhite (approved):
Nice LGTM! 🚢

pgautier404 (approved):

.github/workflows/on-push-to-release-branch.yaml
```diff
@@ -0,0 +1,63 @@
...
+
+      - name: Integration Test
+        run: |
+          dotnet test tests/Integration/Momento.Sdk.Tests
+        shell: bash
+
+      - name: Incubating Integration Test
+        run: |
+          dotnet test tests/Integration/Momento.Sdk.Incubating.Tests
+        shell: bash
```
malandis, line 43: Not a blocker but is there an easy way in GH actions to dry duplicated steps, eg this block vs what's in ci.yaml

malandis (approved):

.github/workflows/on-push-to-release-branch.yaml
cprice404: it can be done.  pete did some of it in the examples repo, see the `.github/actions` dir.  I also have some re-usable actions in `standards-and-practices`.

PR #149, opened 2022-09-13
fix: remove value response name and list push length

### LINKED ISSUE #144: fix listremoveall name and add list length to list push responses

malandis:
This fixes two minor issues:
1. The `ListRemoveValue` response had an out of date name. Changed to `CacheListRemoveValue`
2. Added list length to the response for `ListPushFront` and `ListPushBack` 

Closes #144

### DIFF
--- a/src/Momento.Sdk/Incubating/Internal/ScsDataClient.cs
+++ b/src/Momento.Sdk/Incubating/Internal/ScsDataClient.cs
@@ -447,17 +447,17 @@ internal sealed class ScsDataClient : ScsDataClientBase
         return new CacheListFetchResponse(response);
     }
 
-    public async Task<CacheListRemoveAllResponse> ListRemoveValueAsync(string cacheName, string listName, byte[] value)
+    public async Task<CacheListRemoveValueResponse> ListRemoveValueAsync(string cacheName, string listName, byte[] value)
     {
         return await ListRemoveValueAsync(cacheName, listName, value.ToByteString());
     }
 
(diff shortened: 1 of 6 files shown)


src/Momento.Sdk/Incubating/Internal/ScsDataClient.cs
```diff
@@ -388,15 +386,17 @@ public async Task SendListPushBackAsync(string cacheName, string listName, ByteS
...
 
         try
         {
-            await this.grpcManager.Client.ListPushBackAsync(request, MetadataWithCache(cacheName), deadline: CalculateDeadline());
+            response = await this.grpcManager.Client.ListPushBackAsync(request, MetadataWithCache(cacheName), deadline: CalculateDeadline());
         }
         catch (Exception e)
         {
             throw CacheExceptionMapper.Convert(e);
         }
```
eaddingtonwhite, line 398: maybe null check on response before returning? idk if desired or not just calling out incase.

eaddingtonwhite (approved):

src/Momento.Sdk/Incubating/Internal/ScsDataClient.cs
malandis: This will throw an exception before returning null

eaddingtonwhite: ok sounds good! bit hard to see that from the outside on GH but sounds good.

PR #148, opened 2022-09-13
feat: copy dotnet examples from examples repo

cprice404:
We are in the process of moving the examples out of the old examples repo and into the individual SDK repos.  This commit copies the dotnet examples over; no changes.

Also copies over Pete's dependabot/examples CI work.

PR #154, opened 2022-09-15
fix: add missing github email action

malandis:
This email action was missing from PR #148.


cprice404:
@malandis i think we can close this

PR #157, opened 2022-09-16
feat: split out incubating package

pgautier404:
Splits incubating code out of the SDK. It will live in a new incubating package, which provides a superset of SDK functionality.


cprice404 (dismissed):
this lgtm but a couple of things:

1. @malandis may need to do a release for zigza to get the dictionary increment stuff out to them, and we may want to do that before this is merged, so let's coordinate with him
2. when you create the new git repo let's add it to the notifications in the open-source slack channel.

cprice404 (approved):

PR #163, opened 2022-09-19
feat: flatten exception hierarchy

pgautier404:
Removed `ClientSdkException` and `MomentoServiceException` from the exception hierarchy and modified all subclasses to subclass `SdkException` instead. Added an `UnknownException` type to handle the fallthrough case in `CacheExceptionMapper`.


cprice404 (approved):
:ty:

PR #167, opened 2022-09-20
chore: add .net version in agent header

### LINKED ISSUE #156: include .net version in agent header

poppoerika:
.NET version was already included in agent header, but I updated the agent string from `csharp:` to `dotnet:` and version API.
I added a chart, SDK Agent/Version, on the bottom of this dashboard: https://app.lightstep.com/momento-preprod/dashboard/mr2-overview-do-not-touch/bb7bWC2V?time_window=minutes_60&sort=sdk_type_version.desc
You can observe `cutomer_id = chris+taco@momentohq.com` and `sdk_type_version = dotnet:6.0.9`

Closes #156


malandis:
We need to track two versions: (1) SDK version number and (2) .NET runtime version. I believe the previous commit was tracking the SDK version number (1). We still need to do that in addition to what you've added here, which addresses (2).

pgautier404 (dismissed):

eaddingtonwhite (dismissed):

poppoerika:
@malandis 
Ahh, you're right. What we had before was our SDK version (assembly version) and what we have now is .NET Core version (6.0.9).
If I remember correctly, we first added this header to keep track of SDK version (not language or framework version). 
So other SDKs only have SDK version in their headers. 
I don't mind including .NET Core version in addition to SDK version but I think we want to be consistent throughout the SDKs.
cc @eaddingtonwhite @cprice404 @eaddingtonwhite

cprice404:
@poppoerika yes, @malandis is right, we want both.  And you are right that we will want to make this consistent across the SDKs but we can tackle that after we get dotnet SDK to 1.0.

So I think we need to make a call on whether we add this as a second header, or just append it to the value string for the existing header.  I can start a thread in engineering about that.

poppoerika:
As per the conversations [here](https://gomomento.slack.com/archives/C01AC7JT370/p1663705006957619), I added the second header `Runtime_Version` to the header (indicating .NET/.NET Core version) and also kept `Agent` header as it was (indicating SDK version).

pgautier404 (approved):

malandis (approved):
Thank you!

PR #171, opened 2022-09-22
feat: return errors in responses

pgautier404:


examples/MomentoApplication/Program.cs
```diff
@@ -17,25 +17,50 @@ class Program
...
+                Console.WriteLine($"Exception msg: {error.Exception.Message}");
             }
-            catch (AlreadyExistsException)
-            {
-                Console.WriteLine($"Cache with name {CACHE_NAME} already exists.\n");
+
+            Console.WriteLine($"Creating cache {CACHE_NAME}");
+            CreateCacheResponse createResp = client.CreateCache(CACHE_NAME);
+            if (createResp is CreateCacheResponse.Error errorResp2) {
+                if (errorResp2.Exception.ErrorCode == MomentoErrorCode.ALREADY_EXISTS_ERROR) {
```
cprice404, line 31: looking at this - do you think it would be possible to move the `ErrorCode` up one level?  Just thinking about how I'd *like* to write this, as a user:

```
if (createResp is CreateCacheResponse.Error) {
   if (createResp.ErrorCode == MomentoErrorCode.ALREADY_EXISTS_ERROR) {
      ...
   }
}
```

Also I'm guessing from your code here that if you want the C# compiler to do the smart cast then you have to give it a new variable name as you did with `errorResp2` above?  How did that feel to you when coding?  I am curious if that's a common idiom for C# devs (I honestly have no idea :) )

examples/MomentoApplication/Program.cs
```diff
@@ -17,25 +17,50 @@ class Program
...
+
+            Console.WriteLine("Listing caches");
             String token = null;
-            do
-            {
-                ListCachesResponse resp = client.ListCaches(token);
-                foreach (CacheInfo cacheInfo in resp.Caches)
-                {
-                    Console.WriteLine(cacheInfo.Name);
+            do {
```
cprice404, line 40: @kvcache - pinging you because you're probably the most familiar with C# idioms out of all of our devs.  WDYT about this?  looks like we have to provide a second variable name in order for the smart cast to work.

cprice404 (commented):
i have the one nit about the depth of `ErrorCode`, and the callout that we will have to split out the example changes into a separate PR before we can merge this.  and would like to get @kvcache 's input on it.  but overall i think it looks good!

src/Momento.Sdk/Responses/CreateCacheResponse.cs
```diff
@@ -1,8 +1,23 @@
 ﻿namespace Momento.Sdk.Responses;
 
-public class CreateCacheResponse
-{
-    public CreateCacheResponse()
+using Momento.Sdk.Exceptions;
+
+public class CreateCacheResponse {
```
cprice404, line 5: we can make this abstract now so that it can't be instantiated directly.  not a huge deal but might as well.

examples/MomentoApplication/Program.cs
kvcache: I do not like arbitr abbrevia in variab nam.

Aside from that, I'm not broadly experienced with real modern c# best practices. I know Java has a thing like this too as does rust - seems like a plausible thing to promote but I haven't solicited reviews from a hundred professionals or anything like that. It seems nice to me. I like rebinding the name.

cprice404: yeah we can nitpick about var name choices later, for now i'm just looking for feedback on the API.  I take it you like the rebinding pattern here better than what we had originally proposed where you had to check the enum and then call `.asHit()` etc.  Correct?

cprice404: @gautamomento curious if you have any input on this too.

kvcache: Very much yes I prefer the rebind versus enum check + our own reimplementation and rename of a single-purpose typecast!

I'm seeing green lines in here and it's an "example." I think it's fair to ask whether we want to invent aliases for words like "response" as those new inventions are being proposed here. I read this on my phone and it was unclear where you were drawing attention to (a `do {`?) so I scanned around. Also, the expression in question I think you're looking for feedback on, `if (listResp is ListCachesResponse.Success listSuccessResp)`, is unnecessarily troublesome to parse for meaning as a consequence of abbreviation. Whole word(s) would make this read more naturally. I don't consider this a nit.

cprice404: duly noted.

gautamomento: I prefer the rebinding pattern over enum checks, I find it much easier to read and handy to write.

src/Momento.Sdk/Exceptions/SdkException.cs
```diff
@@ -4,21 +4,21 @@
...
+  PERMISSION_ERROR,  // Insufficient permissions to perform operation
+  AUTHENTICATION_ERROR,  // Invalid authentication credentials to connect to cache service
+  CANCELLED_ERROR,  // Request was cancelled by the server
+  LIMIT_EXCEEDED_ERROR,  // Request rate exceeded the limits for the account
+  BAD_REQUEST_ERROR,  // Request was invalid
+  TIMEOUT_ERROR,  // Client's configured timeout was exceeded
+  SERVER_UNAVAILABLE,  // Serer was unable to handle the request
+  CLIENT_RESOURCE_EXHAUSTED,  // A client resource (most likely memory) was exhausted
+  FAILED_PRECONDITION_ERROR,  // System is not in a state required for the operation's execution
+  UNKNOWN_ERROR  // Unknown error has occurred
```
malandis, line 21: Please use `///` for each comment above each enum value. The IDE will generate the appropriate documentation tag and it will be available in the IDE and generated documentation, eg:

```csharp
/// <summary>
/// Invalid argument passed to Momento client
/// </summary>
INVALID_ARGUMENT_ERROR,
```

malandis (commented):
Minor point about documentation

examples/MomentoApplication/Program.cs
pgautier404: The error code is added into the mix on a per-exception-type basis when the `SdkException` subclasses are constructed, so I left that logic as-is and added a getter to the response object. The getter just grabs and returns the code from the exception.

Fixed in a6184ba5

I'm still playing with the example code, but I did seem to have to use new names every time I did one of those casts. It was a little annoying, but there may be workarounds and I think the extra work is worth it for the ease of matching and casting.

src/Momento.Sdk/Responses/CreateCacheResponse.cs
pgautier404: Fixed in a6184ba5

src/Momento.Sdk/Exceptions/SdkException.cs
pgautier404: Awesome, thanks! Fixed in 3e59371b.

cprice404 (approved):
✨

Issue #78, opened 2022-07-14
c# sdk configurable tuning


cprice404:
- [x] Sketch data structures
- [x] move data client deadline to grpc config
- [x] Add logging
- [ ] Implement retries

PR #172, opened 2022-09-22
chore: move data client deadline to grpc configuration

malandis:
Moves data client gRPC deadline to gRPC configuration.

Work towards #78

### DIFF
--- a/src/Momento.Sdk/Config/Configuration.cs
+++ b/src/Momento.Sdk/Config/Configuration.cs
@@ -7,9 +7,9 @@ namespace Momento.Sdk.Config;
 
 public class Configuration : IConfiguration
 {
-    public IRetryStrategy RetryStrategy { get; set; }
-    public IList<IMiddleware> Middlewares { get; set; }
-    public ITransportStrategy TransportStrategy { get; set; }
+    public IRetryStrategy RetryStrategy { get; }
+    public IList<IMiddleware> Middlewares { get; }
+    public ITransportStrategy TransportStrategy { get; }
 
     public Configuration(IRetryStrategy retryStrategy, ITransportStrategy transportStrategy)
         : this(retryStrategy, new List<IMiddleware>(), transportStrategy)
@@ -22,5 +22,20 @@ public class Configuration : IConfiguration
         this.RetryStrategy = retryStrategy;
         this.Middlewares = middlewares;
         this.TransportStrategy = transportStrategy;
+    }
+
+    public IConfiguration WithRetryStrategy(IRetryStrategy retryStrategy)
+    {
+        return new Configuration(retryStrategy, Middlewares, TransportStrategy);
+    }
+
+    public IConfiguration WithMiddlewares(IList<IMiddleware> middlewares)
+    {
+        return new Configuration(RetryStrategy, middlewares, TransportStrategy);
+    }
+
+    public IConfiguration WithTransportStrategy(ITransportStrategy transportStrategy)
+    {
+        return new Configuration(RetryStrategy, Middlewares, transportStrategy);
     }
 }
--- a/src/Momento.Sdk/Config/Configuration.cs
+++ b/src/Momento.Sdk/Config/Configuration.cs
@@ -7,9 +7,9 @@ namespace Momento.Sdk.Config;
 
 public class Configuration : IConfiguration
 {
-    public IRetryStrategy RetryStrategy { get; }
-    public IList<IMiddleware> Middlewares { get; }
-    public ITransportStrategy TransportStrategy { get; }
+    public IRetryStrategy RetryStrategy { get; set; }
+    public IList<IMiddleware> Middlewares { get; set; }
+    public ITransportStrategy TransportStrategy { get; set; }
 
     public Configuration(IRetryStrategy retryStrategy, ITransportStrategy transportStrategy)
         : this(retryStrategy, new List<IMiddleware>(), transportStrategy)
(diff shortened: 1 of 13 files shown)


src/Momento.Sdk/Config/Configuration.cs
```diff
@@ -7,9 +7,9 @@ namespace Momento.Sdk.Config;
 
 public class Configuration : IConfiguration
 {
-    public IRetryStrategy RetryStrategy { get; }
-    public IList<IMiddleware> Middlewares { get; }
-    public ITransportStrategy TransportStrategy { get; }
+    public IRetryStrategy RetryStrategy { get; set; }
+    public IList<IMiddleware> Middlewares { get; set; }
+    public ITransportStrategy TransportStrategy { get; set; }
```
cprice404, line 12: I'm not sure how strongly I feel about it but my inclination is to disallow setters here.  I would prefer for these config objects to be immutable, otherwise they could theoretically be changed at runtime.

See my middleware PR for an example of a `With` copy constructor that can be used to augment an existing config before it is passed to the client constructor.  Happy to discuss if you disagree.

src/Momento.Sdk/Config/Retry/FixedCountRetryStrategy.cs
```diff
@@ -2,7 +2,7 @@ namespace Momento.Sdk.Config.Retry;
 
 public class FixedCountRetryStrategy : IRetryStrategy
 {
-    public int MaxAttempts { get; private set; }
+    public int MaxAttempts { get; set; }
```
cprice404, line 5: same here

src/Momento.Sdk/Config/Transport/IGrpcConfiguration.cs
```diff
@@ -11,16 +11,21 @@ namespace Momento.Sdk.Config.Transport;
 /// </summary>
 public interface IGrpcConfiguration
 {
-    public int NumChannels { get; }
+    public int NumChannels { get; set; }
 
-    public int MaxSessionMemory { get; }
-    public bool UseLocalSubChannelPool { get; }
+    public int MaxSessionMemory { get; set; }
+    public bool UseLocalSubChannelPool { get; set; }
```
cprice404, line 17: and on all of these

src/Momento.Sdk/Config/Transport/StaticTransportStrategy.cs
```diff
@@ -1,29 +1,44 @@
...
+            {
+                throw new ArgumentException($"Deadline must be strictly positive. Value was: {value}", "DeadlineMilliseconds");
+            }
+            _deadlineMilliseconds = value;
+        }
+    }
+    public IDictionary<string, string> GrpcChannelConfig { get; set; }
 
-    public StaticGrpcConfiguration(int numChannels, int maxSessionMemory, bool useLocalSubChannelPool)
+    public StaticGrpcConfiguration(int numChannels, int maxSessionMemory, bool useLocalSubChannelPool, uint deadlineMilliseconds = 5000)
```
cprice404, line 28: let's make this required and not provide a default

cprice404 (commented):
looks good other than my nits about immutability and not providing a default value.  I agree with the spot you plumbed it in to.

src/Momento.Sdk/Config/Configuration.cs
malandis: I'll revert to immutable and add more builder-like constructs. Let's loop back at a later date to check compatibility with the .NET configuration interfaces

malandis: I changed this in https://github.com/momentohq/client-sdk-dotnet/pull/172/commits/d7e44627336915bb150c211df7b5d4314acd825b

Because` we depend on interfaces in the code, I had to push the `With*` builder methods into the interface. Let me know what you think.

src/Momento.Sdk/Config/Transport/StaticTransportStrategy.cs
malandis: Done

cprice404 (approved):

src/Momento.Sdk/Config/Configurations.cs
```diff
@@ -27,7 +29,7 @@ public static Laptop Latest
                 IRetryStrategy retryStrategy = new FixedCountRetryStrategy(maxAttempts: 3);
                 ITransportStrategy transportStrategy = new StaticTransportStrategy(
                     maxConcurrentRequests: 1,
-                    grpcConfig: new StaticGrpcConfiguration(numChannels: 6, maxSessionMemory: 128, useLocalSubChannelPool: true));
+                    grpcConfig: new StaticGrpcConfiguration(numChannels: 6, maxSessionMemory: 128, useLocalSubChannelPool: true, deadlineMilliseconds: DEFAULT_DEADLINE_MILLISECONDS));
```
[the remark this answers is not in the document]
pgautier404: Removing the default sounds like a good idea to me.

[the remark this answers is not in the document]
eaddingtonwhite: Yup I agree. I think good to force platform dev making prebuild to always be specific/explicit in values and most users should never have to deal with this. Maybe could link to commend somewhere in docs for someone developing middle ware on recomended values? Can be follow up though when someone actually coming in and contributing a new one.

eaddingtonwhite (approved):

PR #173, opened 2022-09-23
feat: Initial sketch of middleware implementation

cprice404:
This commit introduces a proposed implementation for our Middleware.

It is an attempt to expose the most relevant parts of the gRPC Interceptor in a way that is a bit more approachable for end users who do not want to become gRPC experts.  It still (by necessity) exposes some of the gRPC constructs (such as the Metadata container that stores headers, and the gRPC status code), but it abstracts away some of the more obtuse constructs in the interceptor (such as AsyncUnaryCall).

As implemented, this only works for AsyncUnary requests, but we should not currently be making any other types of requests on the data plane.  If we need to support blocking requests (doubtful) or Streaming requests (more plausible), and we want to support middleware on those, we will need to extend the API.

In subsequent commits this can be used to implement retries, and to limit the maximum number of concurrent requests.  In the future it can be used for logging client-side performance data.


src/Momento.Sdk/Config/Middleware/LoggingMiddleware.cs
```diff
@@ -0,0 +1,28 @@
+﻿using System;
+using Grpc.Core.Interceptors;
+
+namespace Momento.Sdk.Config.Middleware
+{
+    public class LoggingMiddleware : IMiddleware
```
cprice404, line 6: This is just an example; we may delete it when we have better examples.  For as long as we keep it around we should switch it over from Console.WriteLine to use the logger, as soon as Michael lands that PR.

src/Momento.Sdk/Exceptions/SdkException.cs
```diff
@@ -3,99 +3,103 @@
...
-  CLIENT_RESOURCE_EXHAUSTED,
-  /// <summary>
-  /// System is not in a state required for the operation's execution
-  /// </summary>
-  FAILED_PRECONDITION_ERROR,
-  /// <summary>
-  /// Unknown error has occurred
-  /// </summary>
-  UNKNOWN_ERROR
+public enum MomentoErrorCode
```
cprice404, line 6: apologies, this diff comes from `dotnet format`; should go away if you hide whitespace in the diff.  Will try to be more careful about doing `dotnet format` as separate PRs going forward.

src/Momento.Sdk/Momento.Sdk.csproj
```diff
@@ -2,7 +2,7 @@
 <Project Sdk="Microsoft.NET.Sdk">
 	<PropertyGroup>
 		<!-- Build Configuration -->
-		<TargetFramework>netstandard2.0</TargetFramework>
+		<TargetFramework>netstandard2.1</TargetFramework>
```
cprice404, line 5: @malandis I went ahead and did this now even though I haven't yet introduced the stuff that requires it.

cprice404:
I can't believe how tiny this diff came out relative to how long it took me to figure out how to get this working with the Interceptor API :)

src/Momento.Sdk/Internal/DataGrpcManager.cs
```diff
@@ -18,12 +22,24 @@ public class DataGrpcManager : IDisposable
...
         this.channel = GrpcChannel.ForAddress(url, new GrpcChannelOptions() { Credentials = ChannelCredentials.SecureSsl });
         List<Header> headers = new List<Header> { new Header(name: Header.AuthorizationKey, value: authToken), new Header(name: Header.AgentKey, value: version), new Header(name: Header.RuntimeVersionKey, value: runtimeVersion) };
-        CallInvoker invoker = this.channel.Intercept(new HeaderInterceptor(headers));
+
+
+        CallInvoker invokerWithMiddlewares = middlewares.Aggregate(
+            this.channel.CreateCallInvoker(),
+            (invoker, middleware) =>
+            {
+                Console.WriteLine($"Adding an interceptor: {middleware}");
```
malandis, line 36: Note to use logging here

src/Momento.Sdk/Internal/DataGrpcManager.cs
```diff
@@ -18,12 +22,24 @@ public class DataGrpcManager : IDisposable
...
+                return invoker.Intercept(new MiddlewareInterceptor(middleware));
+            }
+        );
+
```
malandis, line 40: Will need to make clear in the configuration documentation that the order the middlewares execute will be last-to-first

malandis (commented):
Looks great. Just one small comment about documentation

pgautier404 (approved):

src/Momento.Sdk/Internal/DataGrpcManager.cs
cprice404: good callout.

cprice404:
i'm going to close this branch because it is subsumed by the tweaks that landed in https://github.com/momentohq/client-sdk-dotnet/pull/182

PR #176, opened 2022-09-23
feat: data error response update

pgautier404:


src/Momento.Sdk/Responses/CacheDeleteResponse.cs
```diff
@@ -1,8 +1,27 @@
 ﻿namespace Momento.Sdk.Responses;
 
+using Momento.Sdk.Exceptions;
+
 public class CacheDeleteResponse
```
cprice404, line 5: can't remember if we decided we were going to make these abstract, nbd though

src/Momento.Sdk/Responses/CacheGetBatchResponse.cs
```diff
@@ -1,29 +1,72 @@
...
-    }
+        public Success(IEnumerable<CacheGetResponse> responses)
+        {
+            this.Responses = new(responses);
+        }
 
-    public IEnumerable<string?> Strings()
-    {
-        return Responses.Select(response => response.String());
+        public IEnumerable<CacheGetStatus> Status
```
cprice404, line 17: @malandis you should look at this part.  i'm not super familiar with this part of the API but some random thoughts I have are:

1. We have been getting rid of the enums in other places, and I had assumed that would happen for CacheGet too, but it seems like it's not happening here.
2. do we really need 3 separate methods for this?  would it not be okay to just have a single enumerable that contains the polymorphic parent type?
3. in the loops below it looks like we are assuming that all of the responses are `Success`, but I don't think there is any guarantee that that is actually true?

src/Momento.Sdk/Responses/CacheGetResponse.cs
```diff
@@ -1,11 +1,50 @@
...
 namespace Momento.Sdk.Responses;
 
-public class CacheGetResponse : CacheGetResponseBase
+public class CacheGetResponse
 {
-    public CacheGetResponse(_GetResponse response) : base(response.Result, response.CacheBody)
+    public class Success : CacheGetResponse
+    {
+        public CacheGetStatus Status { get; }
+        protected readonly ByteString? value;
```
cprice404, line 11: I would have expected us to have sub classes for `Hit` and `Miss` here, and thus avoid the nullable stuff like we were doing elsewhere.

tests/Integration/Momento.Sdk.Tests/SimpleCacheDataTest.cs
```diff
@@ -38,14 +38,20 @@ public async Task SetAsync_KeyIsByteArrayValueIsByteArray_HappyPath()
         byte[] key = Utils.NewGuidByteArray();
         byte[] value = Utils.NewGuidByteArray();
         await client.SetAsync(cacheName, key, value);
-        byte[]? setValue = (await client.GetAsync(cacheName, key)).ByteArray;
-        Assert.Equal(value, setValue);
+        CacheGetResponse response = await client.GetAsync(cacheName, key);
+        if (response is CacheGetResponse.Success goodResponse) {
+            byte[]? setValue = goodResponse.ByteArray;
+            Assert.Equal(value, setValue);
+        }
```
malandis, line 45: Test should fail if response is not a success. The previous test would have failed if not

src/Momento.Sdk/Responses/CacheGetBatchResponse.cs
malandis: 2. You could. I included methods to parallel the unary `Get`: get the status, get the payload. Since this is a batch operation I also included to get the individual items
3. That's correct for the way this is implemented now. If this were a single RPC then there would only be on `Success`

tests/Integration/Momento.Sdk.Tests/SimpleCacheDataTest.cs
malandis, line 45: The previous tests assumed a `Success`. Otherwise they would throw an exception and the test would fail. In this version if the response is not a success, the test still passes. So we need an explicit check to assert Success

tests/Integration/Momento.Sdk.Tests/SimpleCacheDataTest.cs
```diff
@@ -38,14 +38,20 @@ public async Task SetAsync_KeyIsByteArrayValueIsByteArray_HappyPath()
...
         key = Utils.NewGuidByteArray();
         value = Utils.NewGuidByteArray();
         await client.SetAsync(cacheName, key, value, ttlSeconds: 15);
-        setValue = (await client.GetAsync(cacheName, key)).ByteArray;
-        Assert.Equal(value, setValue);
+        response = await client.GetAsync(cacheName, key);
+        if (response is CacheGetResponse.Success anotherGoodResponse) {
+            byte[]? setValue = anotherGoodResponse.ByteArray;
+            Assert.Equal(value, setValue);
+        }
```
malandis, line 54: Same test comment

tests/Integration/Momento.Sdk.Tests/SimpleCacheDataTest.cs
```diff
@@ -73,14 +79,20 @@ public async Task SetAsync_KeyIsStringValueIsString_HappyPath()
         string key = Utils.NewGuidString();
         string value = Utils.NewGuidString();
         await client.SetAsync(cacheName, key, value);
-        string? setValue = (await client.GetAsync(cacheName, key)).String();
-        Assert.Equal(value, setValue);
+        CacheGetResponse response = await client.GetAsync(cacheName, key);
+        if (response is CacheGetResponse.Success goodResponse) {
+            string? setValue = goodResponse.String();
+            Assert.Equal(value, setValue);
+        }
```
malandis, line 86: Same test comment

tests/Integration/Momento.Sdk.Tests/SimpleCacheDataTest.cs
```diff
@@ -73,14 +79,20 @@ public async Task SetAsync_KeyIsStringValueIsString_HappyPath()
...
         key = Utils.NewGuidString();
         value = Utils.NewGuidString();
         await client.SetAsync(cacheName, key, value, ttlSeconds: 15);
-        setValue = (await client.GetAsync(cacheName, key)).String();
-        Assert.Equal(value, setValue);
+        response = await client.GetAsync(cacheName, key);
+        if (response is CacheGetResponse.Success anotherGoodResponse) {
+            string? setValue = anotherGoodResponse.String();
+            Assert.Equal(value, setValue);
+        }
```
malandis, line 95: Same test comment

tests/Integration/Momento.Sdk.Tests/SimpleCacheDataTest.cs
```diff
@@ -108,14 +120,19 @@ public async Task SetAsync_KeyIsStringValueIsByteArray_HappyPath()
         string key = Utils.NewGuidString();
         byte[] value = Utils.NewGuidByteArray();
         await client.SetAsync(cacheName, key, value);
-        byte[]? setValue = (await client.GetAsync(cacheName, key)).ByteArray;
-        Assert.Equal(value, setValue);
+        CacheGetResponse response = await client.GetAsync(cacheName, key);
+        if (response is CacheGetResponse.Success goodResponse) {
+            byte[]? setValue = goodResponse.ByteArray;
+            Assert.Equal(value, setValue);
+        }
```
malandis, line 127: Same test comment

tests/Integration/Momento.Sdk.Tests/SimpleCacheDataTest.cs
```diff
@@ -108,14 +120,19 @@ public async Task SetAsync_KeyIsStringValueIsByteArray_HappyPath()
...
 
         key = Utils.NewGuidString();
         value = Utils.NewGuidByteArray();
         await client.SetAsync(cacheName, key, value, ttlSeconds: 15);
-        setValue = (await client.GetAsync(cacheName, key)).ByteArray;
-        Assert.Equal(value, setValue);
+        response = await client.GetAsync(cacheName, key);
+        var anotherGoodResponse = (CacheGetResponse.Success)response;
+        byte[]? anotherSetValue = anotherGoodResponse.ByteArray;
+        Assert.Equal(value, anotherSetValue);
```
malandis, line 135: Same test comment

tests/Integration/Momento.Sdk.Tests/SimpleCacheDataTest.cs
```diff
@@ -180,7 +199,10 @@ public async Task GetBatchAsync_Failure()
         // Set very small timeout for dataClientOperationTimeoutMilliseconds
         using SimpleCacheClient simpleCacheClient = new SimpleCacheClient(Configurations.Laptop.Latest, authToken, DefaultTtlSeconds, 1);
         List<string> keys = new() { Utils.NewGuidString(), Utils.NewGuidString(), Utils.NewGuidString(), Utils.NewGuidString() };
-        await Assert.ThrowsAsync<Momento.Sdk.Exceptions.TimeoutException>(async () => await simpleCacheClient.GetBatchAsync(cacheName, keys));
+        CacheGetBatchResponse response = await simpleCacheClient.GetBatchAsync(cacheName, keys);
+        if (response is CacheGetBatchResponse.Error badResponse) {
+            Assert.Equal(MomentoErrorCode.TIMEOUT_ERROR, badResponse.ErrorCode);
+        }
```
malandis, line 205: Same test comment

src/Momento.Sdk/Responses/CacheDeleteResponse.cs
pgautier404: We did, and I did for the control responses but not the data ones. Fixed in efab849.

src/Momento.Sdk/Responses/CacheGetResponse.cs
pgautier404: You're totally right . . . I was just in a rhythm :-) Fixed in 0aa3060.

tests/Integration/Momento.Sdk.Tests/SimpleCacheDataTest.cs
pgautier404: I noticed that I was potentially causing false positives in those spots too, thanks. I've updated these to use a direct cast outside of an `if`, so the tests will fail if the subtype we're expecting isn't the one we get. This should address the remainder of your comments as well.

src/Momento.Sdk/Responses/CacheGetBatchResponse.cs
```diff
@@ -1,29 +1,84 @@
...
+        {
+            get
+            {
+                var ret = new List<CacheGetStatus>();
+                foreach (CacheGetResponse response in Responses) {
+                    if (response is CacheGetResponse.Hit hitResponse) {
```
cprice404, line 23: These look better/safer now, so thanks for changing them.

I'm still feeling iffy about this part of the API as a whole, so I think we should revisit it and have a convo about it later this week, but I think it's okay for now for this PR.

src/Momento.Sdk/Responses/CacheGetResponse.cs
cprice404: That commit looks a great improvement but the field is still nullable here... it shouldn't need to be anymore, should it?

tests/Unit/Momento.Sdk.Tests/Responses/CacheGetResponseTest.cs
```diff
@@ -14,16 +14,15 @@ public void CorrectResultMapping()
...
-        Assert.Equal(CacheGetStatus.MISS, responseMiss.Status);
-
-        _GetResponse serverResponseBadRequest = new _GetResponse() { Result = ECacheResult.Invalid };
-        _ = Assert.Throws<InternalServerException>(() => new CacheGetResponse(serverResponseBadRequest));
+        // _GetResponse serverResponseMiss = new _GetResponse() { Result = ECacheResult.Miss };
+        // CacheGetResponse.Miss responseMiss = new CacheGetResponse.Miss(serverResponseMiss);
+        // Assert.Equal(CacheGetStatus.MISS, responseMiss.Status);
 
+        // _GetResponse serverResponseBadRequest = new _GetResponse() { Result = ECacheResult.Invalid };
+        // _ = Assert.Throws<InternalServerException>(() => new CacheGetResponse.Success(serverResponseBadRequest));
```
cprice404, line 26: intentional?

cprice404 (commented):
This looks really close to being done IMO, but I'd like to hear you thoughts on whether the bytearray in the Hit class still needs to be nullable.

src/Momento.Sdk/Responses/CacheGetBatchResponse.cs
pgautier404: Agreed, the batch stuff needs some love and we should take a closer look when time allows.

tests/Unit/Momento.Sdk.Tests/Responses/CacheGetResponseTest.cs
pgautier404: Yes, those are obsolete due to class restructuring. The response subtypes are tested in the integration tests, so we're not losing coverage.

pgautier404:
Great callout on `value` no longer needing to be nullable. Fixed in 5b620a0a.

src/Momento.Sdk/Responses/CacheGetResponse.cs
```diff
@@ -1,11 +1,52 @@
...
 namespace Momento.Sdk.Responses;
 
-public class CacheGetResponse : CacheGetResponseBase
+public abstract class CacheGetResponse
 {
-    public CacheGetResponse(_GetResponse response) : base(response.Result, response.CacheBody)
+    public class Hit : CacheGetResponse
+    {
+        public CacheGetStatus Status { get => CacheGetStatus.HIT; }
+        protected readonly ByteString value;
```
cprice404, line 11: cool, thanks.  I think we should merge this as-is because it's a good step forward, but would like to circle back quickly and see if the enum is still necessary at all.  It feels like the only thing that is currently forcing us to keep it is that batch api, and if that's true, we should talk about options for getting rid of that bit, because having the enum here feels inconsistent since we got rid of it in the other places.

cprice404 (approved):
cool, thanks.  I think we should merge this as-is because it's a good step forward, but would like to circle back quickly and see if the enum is still necessary at all.  It feels like the only thing that is currently forcing us to keep it is that batch api, and if that's true, we should talk about options for getting rid of that bit, because having the enum here feels inconsistent since we got rid of it in the other places.

src/Momento.Sdk/Responses/CacheGetResponse.cs
pgautier404: That sounds good. The enum shouldn't be necessary but removing it altogether as part of this effort seemed risky.

PR #177, opened 2022-09-23
chore: improvements to load generator

cprice404:
This makes the following improvements to the load generator:

* Handles throttles without crashing
* Handles timeouts without crashing
* Adds a very simplistic implementation of a "max requests per second" setting,
  which can be used to rate-limit the load generator on the client side


malandis (approved):
Ensures `max_requests_per_second` by introducing a delay when `request_turnaround_time < expected_turnaround_time_at_max_requests_per_second`

PR #181, opened 2022-09-24
feat: add logger to client

malandis:
This PR allows the cache client to accept a `ILoggerFactory` parameter. This is used to create loggers in the cache client itself, any contained objects, and so on.

We allow the `loggerFactory` to be `null`. In this case we use the `NullLoggerFactory` singleton to create concrete loggers. Those loggers will just perform no-ops.


pgautier404 (dismissed):

cprice404 (dismissed):
lgtm; i agree that we should have a default for this argument.

cprice404:
Rebased this to deal with some conflicts to the constructor signatures.

eaddingtonwhite (approved):
🚢

pgautier404 (approved):

PR #182, opened 2022-09-26
feat: max concurrent request middleware

cprice404:
During tuning, I discovered that the .NET grpc library behaves differently than the others I'd previously tested w/rt connections opened to the server.  Specifically, it will detect whenever your number of concurrent requests exceeds a multiple of 100 and open a new connection to the server when it does.  Thus, in my loadgen scenario where I was executing 5000 concurrent requests, the library would end up opening on the order of 50 connections to the server.

This behavior is undesirable for both the client and server, depending on the network bandwidth available between the two.  Specifically, in a laptop environment, throughput and latency both decrease when you exceed about 200 concurrent requests, so there is no value in opening so many additional connections to the server and this could cause server-side throttling/load shedding.

In this commit we introduce a new middleware that can restrict the maximum number of concurrent requests.  Setting this to 200 in the laptop environment results in far fewer connections, and throughput and latency for the happy path cases are unaffected. For the degenerate case (5000+ concurrent requests), this protects the server and actually seems to improve client-side p999 latencies by quite a bit.

Note that as part of this commit I needed to re-work the middleware layer.  We can't rely directly on using gRPC Interceptors for this, because their method signatures are not asynchronous.  Thus, we would end up blocking threads to wait for any kind of semaphore, which would have a big negative impact on performance.  Instead, in this commit, we move the middleware layer to some code that we own ourselves so that we are able to use async method signatures.

### DIFF
--- a/src/Momento.Sdk/Config/Configuration.cs
+++ b/src/Momento.Sdk/Config/Configuration.cs
@@ -1,4 +1,5 @@
 using System.Collections.Generic;
+using System.Linq;
 using Momento.Sdk.Config.Middleware;
 using Momento.Sdk.Config.Retry;
 using Momento.Sdk.Config.Transport;
@@ -37,5 +38,14 @@ public class Configuration : IConfiguration
     public IConfiguration WithTransportStrategy(ITransportStrategy transportStrategy)
     {
         return new Configuration(RetryStrategy, Middlewares, transportStrategy);
+    }
+
+    public Configuration WithAdditionalMiddlewares(IList<IMiddleware> additionalMiddlewares)
+    {
+        return new(
+            retryStrategy: RetryStrategy,
+            middlewares: Middlewares.Concat(additionalMiddlewares).ToList(),
+            transportStrategy: TransportStrategy
+        );
     }
 }
--- a/src/Momento.Sdk/Config/Configurations.cs
+++ b/src/Momento.Sdk/Config/Configurations.cs
@@ -28,7 +28,7 @@ public class Configurations
                 /*retryableStatusCodes = DEFAULT_RETRYABLE_STATUS_CODES,*/
                 IRetryStrategy retryStrategy = new FixedCountRetryStrategy(maxAttempts: 3);
                 ITransportStrategy transportStrategy = new StaticTransportStrategy(
-                    maxConcurrentRequests: 1,
+                    maxConcurrentRequests: 200,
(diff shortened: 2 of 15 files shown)


src/Momento.Sdk/Config/Middleware/FairAsyncSemaphore.cs
```diff
@@ -0,0 +1,39 @@
+﻿using System;
+using System.Threading.Channels;
+using System.Threading.Tasks;
+
+namespace Momento.Sdk.Config.Middleware
+{
+    public class FairAsyncSemaphore
```
eaddingtonwhite, line 7: this class seems fine the only thing seems little off is this is in Middleware package? Seems like a class to help with general concurency? Just confused me for second thought this was some other middle ware till I realized it was consumed by rate limited one further down. Not big issue just calling out made me double take for second.

eaddingtonwhite (approved):
This looks good. Nice write up in PR as well! Maybe makes sense to port some of that into comment in middleware so dont loose context as well?  I'm good w/ 🚢 this as is though.

src/Momento.Sdk/Config/Middleware/FairAsyncSemaphore.cs
```diff
@@ -0,0 +1,39 @@
...
+        public async Task WaitOne()
+        {
+            //Console.WriteLine($"Waiting for semaphore");
+            await _ticketChannel.Reader.ReadAsync();
+        }
+
+        public async Task Release()
+        {
+            //Console.WriteLine($"Releasing for semaphore");
+            await _ticketChannel.Writer.WriteAsync(true);
```
kvcache, line 35:
```suggestion
        public void Release()
        {
            var balanced = _ticketChannel.Writer.TryWrite(true);
            if (!balanced)
            {
                throw new ApplicationException("more releases than waits! These must be 1:1")
            }
```

if a user were to put in imbalanced releases then their coroutines will all stack up in memory and it will eventually OOM. If you do this instead then it still won't necessarily show the offending stack but it'll at least tell you promptly that something is wrong.

Also if you do it this way there's no task and state machine so it'll be a handful of µs faster

kvcache (commented):
looks fine. 1 quality of life suggestion for your users

src/Momento.Sdk/Config/Middleware/FairAsyncSemaphore.cs
cprice404: good call, ty

cprice404:
rebased, which required a fair number of tweaks to constructor signagures.

eaddingtonwhite (approved):

PR #184, opened 2022-09-26
fix: exceptions updates

pgautier404:


src/Momento.Sdk/Exceptions/AlreadyExistsException.cs
```diff
@@ -1,14 +1,14 @@
...
 /// <summary>
 /// Resource already exists
 /// </summary>
 public class AlreadyExistsException : SdkException
 {
-    public AlreadyExistsException(string message) : base(MomentoErrorCode.ALREADY_EXISTS_ERROR, message)
-    {
-    }
-    public AlreadyExistsException(string message, MomentoErrorTransportDetails transportDetails) : base(MomentoErrorCode.ALREADY_EXISTS_ERROR, message, transportDetails)
+    public AlreadyExistsException(string message, MomentoErrorTransportDetails? transportDetails=null, Exception? e=null) : base(MomentoErrorCode.ALREADY_EXISTS_ERROR, message, transportDetails, e)
```
cprice404, line 10: curious what prompted the change to nullable for the transport details here.  not a blocker.

src/Momento.Sdk/Exceptions/BadRequestException.cs
```diff
@@ -1,14 +1,14 @@
...
 /// </summary>
 public class BadRequestException : SdkException
 {
-    public BadRequestException(string message) : base(MomentoErrorCode.BAD_REQUEST_ERROR, message)
-    {
-    }
-    public BadRequestException(string message, MomentoErrorTransportDetails transportDetails) : base(MomentoErrorCode.BAD_REQUEST_ERROR, message, transportDetails)
+    public BadRequestException(string message, MomentoErrorTransportDetails? transportDetails=null, Exception? e=null) : base(MomentoErrorCode.BAD_REQUEST_ERROR, message, transportDetails, e)
     {
+        this.MessageWrapper = "The request was invalid; please contact Momento";
```
cprice404, line 12: for this one in particular, things may have changed a bit since the time i wrote the spec.  For this one we are definitely expecting the gRPC error from the server to contain more information about why the request was invalid, so we want to try to make sure we surface that as part of the most visible user-facing string.  So perhaps we want to interpolate it into this string?

src/Momento.Sdk/Exceptions/CancelledException.cs
```diff
@@ -1,14 +1,14 @@
...
     /// <summary>
     /// Operation was cancelled.
     /// </summary>
-    public CancelledException(string message) : base(MomentoErrorCode.CANCELLED_ERROR, message)
-    {
-    }
-    public CancelledException(string message, MomentoErrorTransportDetails transportDetails) : base(MomentoErrorCode.CANCELLED_ERROR, message, transportDetails)
+    public CancelledException(string message, MomentoErrorTransportDetails? transportDetails=null, Exception? e=null) : base(MomentoErrorCode.CANCELLED_ERROR, message, transportDetails)
     {
+        this.MessageWrapper = "The request was cancelled by the server; please contact Momento";
```
cprice404, line 12: in all the places where I wrote "Please contact Momento" I think the latest suggestion was to change it to something like: "for assistance, please contact us at support@momentohq.com".

src/Momento.Sdk/Exceptions/FailedPreconditionException.cs
```diff
@@ -8,10 +10,8 @@ namespace Momento.Sdk.Exceptions;
 /// </summary>
 public class FailedPreconditionException : SdkException
 {
-    public FailedPreconditionException(string message) : base(MomentoErrorCode.FAILED_PRECONDITION_ERROR, message)
-    {
-    }
-    public FailedPreconditionException(string message, MomentoErrorTransportDetails transportDetails) : base(MomentoErrorCode.FAILED_PRECONDITION_ERROR, message, transportDetails)
+    public FailedPreconditionException(string message, MomentoErrorTransportDetails? transportDetails=null, Exception? e=null) : base(MomentoErrorCode.FAILED_PRECONDITION_ERROR, message, transportDetails, e)
     {
+        this.MessageWrapper = "System is not in a state required for the operation's execution";
```
cprice404, line 15: I think we want to pass the server message through on this one too.  AFAIK the only place on the server where we are currently using this is if they try to do an "increment" on a non-string value and the message from the server should hopefully state that pretty clearly.

src/Momento.Sdk/Responses/CacheDeleteResponse.cs
```diff
@@ -23,5 +23,9 @@ public MomentoErrorCode ErrorCode
         {
             get => _error.ErrorCode;
         }
+
+        public string Message {
+            get => $"{_error.MessageWrapper}: {_error.Message}";
```
cprice404, line 28: oh, I see - so with this pattern we would always be surfacing both your wrapper message and the one from the server?

cprice404 (commented):
one question about how the server error gets surfaced, and one comment about including "support@momentohq.com" in the strings where we suggest contacting momento, otherwise i think lgtm

src/Momento.Sdk/Exceptions/AlreadyExistsException.cs
pgautier404: There's a fallthrough case in `CacheExceptionMapper` where we don't have transport details to add (`UnknownException`).

src/Momento.Sdk/Responses/CacheDeleteResponse.cs
pgautier404: Yes, it includes the static preamble for context and then the specific exception message coming from the server (or validators or whatever).

src/Momento.Sdk/Exceptions/AlreadyExistsException.cs
cprice404: but wouldn't that only impact the signature for the `UnknownException` constructor, and not these other ones?

cprice404 (approved):

src/Momento.Sdk/Exceptions/InternalServerException.cs
```diff
@@ -7,13 +7,8 @@ namespace Momento.Sdk.Exceptions;
...
-    public InternalServerException(string message) : base(MomentoErrorCode.INTERNAL_SERVER_ERROR, message)
-    {
-    }
-    public InternalServerException(string message, Exception e) : base(MomentoErrorCode.INTERNAL_SERVER_ERROR, message, null, e)
-    {
-    }
-    public InternalServerException(string message, MomentoErrorTransportDetails transportDetails, Exception e) : base(MomentoErrorCode.INTERNAL_SERVER_ERROR, message, transportDetails, e)
+    public InternalServerException(string message, MomentoErrorTransportDetails? transportDetails=null, Exception? e=null) : base(MomentoErrorCode.INTERNAL_SERVER_ERROR, message, transportDetails, e)
     {
+        this.MessageWrapper = "An unexpected error occurred while trying to fulfill the request; please contact Momento";
```
cprice404, line 12: missed one?

src/Momento.Sdk/Exceptions/LimitExceededException.cs
```diff
@@ -1,14 +1,14 @@
...
 /// </summary>
 public class LimitExceededException : SdkException
 {
-    public LimitExceededException(string message) : base(MomentoErrorCode.LIMIT_EXCEEDED_ERROR, message)
-    {
-    }
-    public LimitExceededException(string message, MomentoErrorTransportDetails transportDetails) : base(MomentoErrorCode.LIMIT_EXCEEDED_ERROR, message, transportDetails)
+    public LimitExceededException(string message, MomentoErrorTransportDetails? transportDetails, Exception? e=null) : base(MomentoErrorCode.LIMIT_EXCEEDED_ERROR, message, transportDetails, e)
     {
+        this.MessageWrapper = "Request rate exceeded the limits for this account.  To resolve this error, reduce your request rate, or contact Momento to request a limit increase";
```
cprice404, line 12: here too

src/Momento.Sdk/Exceptions/InternalServerException.cs
pgautier404: Ack, sorry, I assumed you'd pointed them all out. Fixing these now.

cprice404 (approved):
still a couple of places where we should put that email address but it's fine with me either way, this PR or follow-up

cprice404 (approved):

PR #186, opened 2022-09-26
chore: remove default timeout and set it explicitly from pre-built configs

cprice404:


src/Momento.Sdk/Config/Configurations.cs
```diff
@@ -60,7 +58,8 @@ public static Default Latest
                     IRetryStrategy retryStrategy = new FixedCountRetryStrategy(maxAttempts: 3);
                     ITransportStrategy transportStrategy = new StaticTransportStrategy(
                         maxConcurrentRequests: 1,
-                        grpcConfig: new StaticGrpcConfiguration(numChannels: 6, maxSessionMemory: 128, useLocalSubChannelPool: true, deadlineMilliseconds: DEFAULT_DEADLINE_MILLISECONDS));
+                        // TODO: tune the timeout value
```
malandis, line 61: Agreed with the comment. The timeout becomes an unexpected issue, especially with batch commands,

malandis (approved):

pgautier404 (approved):

PR #193, opened 2022-09-27
chore: update examples to momento SDK 0.30.1

cprice404:


examples/MomentoFSharpApplication/Program.fs
```diff
@@ -19,9 +22,11 @@ let exerciseCache() = (
...
         printfn("Listing caches:")
         let resp = client.ListCaches()
-        for cacheInfo in resp.Caches do
-            printfn $"{cacheInfo.Name}"
-        
+        let _ = match resp with
+                | :? ListCachesResponse.Success as successResponse ->
+                    for cacheInfo in successResponse.Caches do
+                        printfn $"{cacheInfo.Name}"
+                | _ -> ()
```
malandis, line 29: :heart_eyes:

malandis (approved):

examples/MomentoFSharpApplication/Program.fs
poppoerika: Yesterday was my first encounter of F# 😱

examples/MomentoFSharpApplication/Program.fs
```diff
@@ -30,5 +35,5 @@ let exerciseCache() = (
 
 [<EntryPoint>]
 let main(argv :string[]) =
-    exerciseCache()
+    let _ = exerciseCache()
```
poppoerika, line 38: Is `_` similar to Go? (not using the assigned variable?)

examples/MomentoApplication/Program.cs
```diff
@@ -19,14 +19,8 @@ class Program
...
-            try
-            {
-                client.CreateCache(CACHE_NAME);
-            }
-            catch (AlreadyExistsException)
-            {
-                Console.WriteLine($"Cache with name {CACHE_NAME} already exists.\n");
-            }
+            var createCacheResult = client.CreateCache(CACHE_NAME);
+
```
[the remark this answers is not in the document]
cprice404: Correct!  It's part of our overall design shift away from exceptions and towards including the information as part of the return type.  The `createCacheResult` here has the info about whether the cache was created successfully.

Feel free to play around with the example code and see how it feels!  We could definitely add an `if` block after this line if we wanted to make it more obvious.

examples/MomentoFSharpApplication/Program.fs
cprice404: correct.  But f# really wants you to avoid statements in favor of expressions so it was giving me a warning that I wasn't doing anything with the return value.

pgautier404 (approved):

PR #198, opened 2022-09-27
feat: async control commands

### LINKED ISSUE #197: async control operations
The .NET SDK has async methods for data plane operations. Unlike this, the control plane operations are blocking. Because the API should be consistent, change these blocking methods to be async.

malandis:
This PR changes the control plane operations to be async instead of blocking.

Closes #197


cprice404 (approved):
checks out

PR #205, opened 2022-09-28
chore: remove batch operations

pgautier404:


poppoerika (approved):
Looks good to me!

One question: do we want to keep those batch operations in a separate branch since a few customers requested and we can revisit these easily in the future?

pgautier404:
@poppoerika The batch operations have been moved to a branch in client-sdk-dotnet-incubating and will be added into the incubating client in the future.

PR #243, opened 2022-09-29
chore: remove batch response classes that were moved to incubating

pgautier404:
These have been moved to the incubating repository already but were accidentally not deleted from this repo.


poppoerika (approved):
Can we release the SDK so that I can use this version SDK without batch response type?
Or VS is giving me an error sating that batch response exist in both SDKs so it's ambiguous.

PR #246, opened 2022-09-29
fix: update sdk examples

pgautier404:


examples/MomentoApplication/Program.cs
```diff
@@ -6,43 +6,94 @@
...
+    if (createCacheResponse is CreateCacheResponse.Error createError)
     {
-        static readonly String MOMENTO_AUTH_TOKEN = Environment.GetEnvironmentVariable("MOMENTO_AUTH_TOKEN");
-        static readonly String CACHE_NAME = "cache";
-        static readonly String KEY = "MyKey";
-        static readonly String VALUE = "MyData";
-        static readonly uint DEFAULT_TTL_SECONDS = 60;
-
-        async static Task Main(string[] args)
+        if (createError.ErrorCode == MomentoErrorCode.LIMIT_EXCEEDED_ERROR)
```
poppoerika, line 23: nit: maybe `createError` to `createCacheError`?

examples/MomentoApplication/Program.cs
```diff
@@ -6,43 +6,94 @@
...
+        }
+        // Any error is considered fatal.
+        Environment.Exit(1);
+    }
+    // If there's already a cache by this name, we don't want to modify it.
+    if (createCacheResponse is CreateCacheResponse.CacheAlreadyExists)
+    {
+        Console.WriteLine($"Error: strangely, a cache with the name {CACHE_NAME} already exists. Exiting.");
+        Environment.Exit(1);
+    }
```
poppoerika, line 38: maybe I'm misunderstanding how we handle `CacheAlreadyExists` but I thought we want to pass it instead of exiting the program?

examples/MomentoApplication/Program.cs
```diff
@@ -6,43 +6,94 @@
 using Momento.Sdk.Exceptions;
 using Momento.Sdk.Responses;
 
-namespace MomentoApplication
+string MOMENTO_AUTH_TOKEN = Environment.GetEnvironmentVariable("MOMENTO_AUTH_TOKEN");
+// Use a GUID for the cache name so it's unlikely to exist already.
+string CACHE_NAME = Guid.NewGuid().ToString();
```
poppoerika, line 11: do we want to use `readonly string` instead per Kenny's comment?
https://www.notion.so/momentohq/Kenny-notes-cf5dc7b645a34cb094b7ffee69998624#6b61a3c6e2404b3e8593470f1a5260ae

examples/MomentoApplication/Program.cs
pgautier404: For the most part that's true. The intent of this specific example is a little different, though, and this check is preventing the sample code from writing into a preexisting cache. This is really just showing that the response type exists and can be processed and acted upon.

examples/MomentoApplication/Program.cs
```diff
@@ -6,43 +6,94 @@
...
-                    token = successResult.NextPageToken;
-                }
-            } while (!String.IsNullOrEmpty(token));
-            Console.WriteLine($"\nSetting key: {KEY} with value: {VALUE}");
-            await client.SetAsync(CACHE_NAME, KEY, VALUE);
-            Console.WriteLine($"\nGet value for  key: {KEY}");
-            CacheGetResponse getResponse = await client.GetAsync(CACHE_NAME, KEY);
-            if (getResponse is CacheGetResponse.Hit hitResponse)
+    Console.WriteLine("\nListing caches:");
+    String token = null;
```
poppoerika, line 41: What are the differences between `String` and `string` in C#?

examples/MomentoApplication/Program.cs
```diff
@@ -6,43 +6,94 @@
...
+        } else if (response is ListCachesResponse.Error listError)
+        {
+            // We do not consider this a fatal error, so we just report it.
+            Console.WriteLine($"Error listing caches: {listError.Message}");
+            break;
         }
+    } while (!String.IsNullOrEmpty(token));
+
+    Console.WriteLine($"\nSetting key: {KEY} with value: {VALUE}");
+    var setResponse = await client.SetAsync(CACHE_NAME, KEY, VALUE);
```
poppoerika, line 61: If we want to use `setResponse` as a var name, we might want to use in other places as well such as `listResponse` so that it's more coherent.

examples/MomentoApplication/Program.cs
pgautier404: Essentially there is none. `string` is an alias for `System.String`.

examples/MomentoApplication/Program.cs
poppoerika: 👍

examples/MomentoApplication/Program.cs
```diff
@@ -6,43 +6,94 @@
...
+        ListCachesResponse response = await client.ListCachesAsync(token);
+        if (response is ListCachesResponse.Success successResponse)
+        {
+            foreach (CacheInfo cacheInfo in successResponse.Caches)
             {
-                Console.WriteLine($"\nLookedup value: {hitResponse.String()}, Stored value: {VALUE}");
+                Console.WriteLine($"- {cacheInfo.Name}");
             }
+            token = successResponse.NextPageToken;
+        } else if (response is ListCachesResponse.Error listError)
```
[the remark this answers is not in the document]
pgautier404: Good suggestions . . . I've standardized the naming some in a07686e. Let me know what you think!

examples/MomentoApplication/Program.cs
pgautier404: Addressed in a07686e.

poppoerika:
I was commenting each row but I thought it's too much! 
Overall looks good to me. Just variable naming can be a bit more consistenet?
For example:
`createCacheResponse`, `createCacheSuccess`, `createCacheError` for createCache operation.

examples/MomentoApplication/Program.cs
cprice404: I don't think we should uuid here, because we have a default limit of ~5-10 caches per user.  We don't want folks who run this program a few times to end up hitting their limit.  In fact I think I might be inclined to say we should just use the same default cache name here that you get when you first run the momento cli.

examples/MomentoApplication/Program.cs
```diff
@@ -6,43 +6,94 @@
...
         {
-            using SimpleCacheClient client = new SimpleCacheClient(Configurations.Laptop.Latest, MOMENTO_AUTH_TOKEN, DEFAULT_TTL_SECONDS);
-            var createCacheResult = await client.CreateCacheAsync(CACHE_NAME);
+            Console.WriteLine($"Error creating cache: {createCacheError.Message}. Exiting.");
+        }
+        // Any error is considered fatal.
+        Environment.Exit(1);
+    }
+    // If there's already a cache by this name, we don't want to modify it.
+    if (createCacheResponse is CreateCacheResponse.CacheAlreadyExists)
```
cprice404, line 34: I like this, thanks!

examples/MomentoApplication/Program.cs
cprice404: I like the log message but I think that's enough to illustrate how they could handle it, I think I'd lean towards making this not-fatal too

src/Momento.Sdk/Internal/ScsControlClient.cs
```diff
@@ -32,6 +33,9 @@ public async Task<CreateCacheResponse> CreateCacheAsync(string cacheName)
         }
         catch (Exception e)
         {
+            if (e is RpcException ex && ex.StatusCode == StatusCode.AlreadyExists) {
+                return new CreateCacheResponse.CacheAlreadyExists();
```
cprice404, line 37: ❤️

cprice404 (commented):
looks awesome, ty.  I think we should make that AlreadyExists non-fatal but otherwise lgtm!

poppoerika (commented):
Thank you for updating the variable names! Looks good to me.

pgautier404:
@cprice404 I changed the already exists warning a little in [321d413](https://github.com/momentohq/client-sdk-dotnet/pull/246/commits/321d41376da5f2ba47e36f4be24d5b1eee6abe1e) so it feels (hopefully) less like an unhandled error. Let me know what you think.

cprice404 (approved):
✨

examples/MomentoApplication/Program.cs
```diff
@@ -6,43 +6,93 @@
...
+    string token = null;
+    do
+    {
+        ListCachesResponse listCachesResponse = await client.ListCachesAsync(token);
+        if (listCachesResponse is ListCachesResponse.Success listCachesSuccess)
+        {
+            foreach (CacheInfo cacheInfo in listCachesSuccess.Caches)
             {
-                Console.WriteLine($"\nLookedup value: {hitResponse.String()}, Stored value: {VALUE}");
+                Console.WriteLine($"- {cacheInfo.Name}");
```
malandis, line 48: Clean

examples/MomentoUsage/Program.cs
```diff
@@ -2,22 +2,33 @@
 using Momento.Sdk.Config;
 using Momento.Sdk.Responses;
 
-String MOMENTO_AUTH_TOKEN = Environment.GetEnvironmentVariable("MOMENTO_AUTH_TOKEN");
-String CACHE_NAME = "cache";
-String KEY = "MyKey";
-String VALUE = "MyData";
-uint DEFAULT_TTL_SECONDS = 60;
+string MOMENTO_AUTH_TOKEN = Environment.GetEnvironmentVariable("MOMENTO_AUTH_TOKEN") ?? "";
```
malandis, line 5: Do we want a default value here, or to throw an exception right away?

pgautier404: I was planning on letting our improved error handling for auth tokens handle this.

cprice404: i think we should do both.  doesn't have to be in this PR but we should have an error that says "you forgot to set your env var", and another one that says "you passed a null or empty auth token".

cprice404 (approved):
i'm good to ship this but let's not close out those tickets about the auth token until we have an error that says "missing env var" in these examples.

PR #248, opened 2022-09-29
fix: improve invalid jwt error message

pgautier404:


pgautier404:
A follow-on PR will include error handling for a missing MOMENTO_AUTH_TOKEN env var in the example code.

malandis (changes requested):
This is really timely! Will you also add a check and separate error message if the `authToken` is `null`?

malandis (approved):

PR #251, opened 2022-09-30
chore: include debug symbols and sources

### LINKED ISSUE #217: add source code to distributables

malandis:
Modifies the NuGet package to include debug symbols and sources.

Closes #217

### DIFF
--- a/src/Momento.Sdk/Momento.Sdk.csproj
+++ b/src/Momento.Sdk/Momento.Sdk.csproj
@@ -5,8 +5,13 @@
 		<TargetFramework>netstandard2.1</TargetFramework>
 		<LangVersion>latest</LangVersion>
 		<Nullable>enable</Nullable>
+		<!-- Include documentation in build -->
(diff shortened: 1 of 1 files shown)


eaddingtonwhite (approved):
noice! :shipit:  

Just curious is there a way for customer to strip this out after the fact in their build process? Thinking like a c# aws lambda where they trying to get size down of package? No sure if this bloats it alot or not. Trying to think why someone wouldnt want us to do this.

pgautier404 (approved):

malandis:
> noice! :shipit:
> 
> Just curious is there a way for customer to strip this out after the fact in their build process? Thinking like a c# aws lambda where they trying to get size down of package? No sure if this bloats it alot or not. Trying to think why someone wouldnt want us to do this.

I raised the same concern. We agreed that the gRPC dependencies are the long pole in the deploy size. They're much much larger than the debug/symbol info for us, so it doesn't make sense to strip that out.

PR #252, opened 2022-09-30
fix: add ToString overrides where appropriate

pgautier404:


src/Momento.Sdk/Responses/CacheDeleteResponse.cs
```diff
@@ -2,7 +2,7 @@
 
 using Momento.Sdk.Exceptions;
 
-public class CacheDeleteResponse
+public abstract class CacheDeleteResponse
```
malandis, line 5: Why the abstract modifier?

malandis (commented):
Looks good. Just one question

src/Momento.Sdk/Responses/CacheDeleteResponse.cs
pgautier404: All of the parent response classes are supposed to be abstract so they can't be instantiated. I just missed this one on the first go.

malandis: Thanks for the context!

malandis (approved):

PR #255, opened 2022-09-30
fix: improvements to sdk error structure

pgautier404:


src/Momento.Sdk/Exceptions/ServerUnavailableException.cs
```diff
@@ -0,0 +1,14 @@
+namespace Momento.Sdk.Exceptions;
+
+using System;
+
+/// <summary>
+/// Server was unable to handle the request
```
malandis, line 6:
```suggestion
/// Server was unable to handle the request.
```

malandis (changes requested):
Looks good but my copyedit brain saw a missing "."

malandis (approved):

PR #256, opened 2022-10-03
chore: use snupkg format

malandis:
The recommended `SymbolPackageFormat` is `snupkg`. The default is
deprecated. See the note here: https://learn.microsoft.com/en-us/nuget/create-packages/symbol-packages-snupkg


pgautier404 (approved):

cprice404 (commented):
🙈

PR #262, opened 2022-10-04
chore: update load gen to use new error handling

cprice404:
This gets rid of some straggling try/catch blocks in the loadgen and ensures that it will count throttles and other response types properly.

### DIFF
--- a/examples/MomentoLoadGen/Program.cs
+++ b/examples/MomentoLoadGen/Program.cs
@@ -16,7 +16,6 @@ namespace MomentoLoadGen
     public record CsharpLoadGeneratorOptions
     (
         int printStatsEveryNRequests,
-        uint requestTimeoutMs,
         int cacheItemPayloadBytes,
         int numberOfConcurrentRequests,
         int maxRequestsPerSecond,
@@ -27,8 +26,8 @@ namespace MomentoLoadGen
     {
         SUCCESS,
         UNAVAILABLE,
-        DEADLINE_EXCEEDED,
-        RESOURCE_EXHAUSTED,
+        TIMEOUT,
+        LIMIT_EXCEEDED,
         RST_STREAM
     };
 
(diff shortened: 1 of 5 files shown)


examples/MomentoLoadGen/Program.cs
```diff
@@ -237,77 +236,87 @@ private async Task IssueAsyncSetGet(SimpleCacheClient client, CsharpLoadGenerato
...
-                switch (innerException!.StatusCode)
-                {
+                return Tuple.Create<AsyncSetGetResult, CacheGetResponse?>(AsyncSetGetResult.SUCCESS, miss);
+            } else if (getResponse is CacheGetResponse.Error error)
+            {
+                return Tuple.Create<AsyncSetGetResult, CacheGetResponse?>(ConvertErrorToAsyncSetGetResult(error.ErrorCode, error.Exception), null);
+            } else
+            {
+                throw new ApplicationException($"Unsupported get response: {getResponse}");
+            }
```
eaddingtonwhite, line 280: total nit but I prefer switch statement over the multiline if. Not sure if c# reason though for doing in IF is cleaner with how we pass back error codes so not blocker at all. Just personal preference thing for me. Saw we had switch on status before.

examples/MomentoLoadGen/Program.cs
```diff
@@ -360,13 +370,16 @@ internal class Program
...
-                       options.TimestampFormat = "hh:mm:ss ";
-                   }));
+            return LoggerFactory.Create(builder => {
+                builder.AddSimpleConsole(options =>
+                {
+                    options.IncludeScopes = true;
+                    options.SingleLine = true;
+                    options.TimestampFormat = "hh:mm:ss ";
+                });
+                builder.AddFilter("Grpc.Net.Client.Internal.GrpcCall", LogLevel.Error);
```
malandis, line 380: This is really nice now

malandis (approved):

pgautier404 (approved):
LGTM!

examples/MomentoLoadGen/Program.cs
cprice404: Unfortunately the blessed/idiomatic way to do this kind of pattern matching in C# today is with if statements:

https://learn.microsoft.com/en-us/dotnet/csharp/fundamentals/functional/pattern-matching#type-tests

I would rather have it as a switch too, but the smart casting to the local variable is kind of a crucial part of the user experience here now.

eaddingtonwhite: ack figured was something like this.

PR #263, opened 2022-10-04
fix: remove unnecessary deps (e.g. duplicate grpc dep)

cprice404:
This commit attempts to pare down our dependency list to the
minimum requirements to compile.  Previously we had more than
one gRPC library, plus a google protobuf lib that we don't seem
to need.

This also removes the SourceLink.Github because it wasn't required
for compiling, but that is probably a mistake and should be
reverted... just posting this for discussion since I'm not familiar
with that one.

### DIFF
--- a/src/Momento.Sdk/Momento.Sdk.csproj
+++ b/src/Momento.Sdk/Momento.Sdk.csproj
@@ -39,6 +39,8 @@
 	  <None Remove="Momento.Protos" />
 	  <None Remove="JWT" />
 	  <None Remove="System.IdentityModel.Tokens.Jwt" />
+	  <None Remove="Microsoft.Extensions.Logging" />
+	  <None Remove="Microsoft.SourceLink.GitHub" />
 	</ItemGroup>
 	<ItemGroup>
 	  <Folder Include="Internal\Middleware\" />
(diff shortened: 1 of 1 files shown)


src/Momento.Sdk/Momento.Sdk.csproj
```diff
@@ -32,25 +32,22 @@
...
 	</PropertyGroup>
-	<ItemGroup>
-		<PackageReference Include="Google.Protobuf" Version="3.19.0" />
-		<PackageReference Include="Grpc.Net.Client" Version="2.40.0" />
-		<PackageReference Include="Grpc.Core" Version="2.41.1" />
-		<PackageReference Include="Microsoft.Extensions.Logging" Version="6.0.0" />
-		<PackageReference Include="Microsoft.SourceLink.Github" Version="1.1.1">
-		  <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
-		  <PrivateAssets>all</PrivateAssets>
-		</PackageReference>
```
cprice404, line 43: I did all of this through the Visual Studio UI.  I started by removing all deps, then added them back in until it compiled and tests pass.  I suspect that this Sourcelink.Github thing is something we want to keep but the build and tests all pass without it and I'm not completely sure what it is, so wanted to ask.  I assume I should add it back.

pgautier404: @malandis this is what you just added to get the source code bundled with the SDK, yes?

malandis: @pgautier404 @cprice404 Yes I added this today to get the source code linked. It's only used at build time: https://github.com/dotnet/sourcelink

cprice404: @malandis I pushed a commit that adds that back, along with the logging lib.

malandis (approved):

PR #267, opened 2022-10-04
feat: replace authToken with more flexible authProvider

pgautier404:


src/Momento.Sdk/Auth/ICredentialProvider.cs
```diff
@@ -0,0 +1,10 @@
+namespace Momento.Sdk.Auth;
+
+using System.Collections.Generic;
+
+public interface ICredentialProvider
+{
+    public string AuthToken { get; }
+    public string ControlEndpoint {get; }
+    public string CacheEndpoint { get; }
```
kvcache, line 9: public is already the default visibility here on an interface.

src/Momento.Sdk/SimpleCacheClient.cs
```diff
@@ -26,26 +27,17 @@ public class SimpleCacheClient : ISimpleCacheClient
...
     /// <param name="defaultTtlSeconds">Default time to live for the item in cache.</param>
     /// <param name="loggerFactory">Logger factory to create loggers for contained instances.</param>
-    public SimpleCacheClient(IConfiguration config, string authToken, uint defaultTtlSeconds, ILoggerFactory? loggerFactory = null)
+    public SimpleCacheClient(IConfiguration config, ICredentialProvider authProvider, uint defaultTtlSeconds, ILoggerFactory? loggerFactory = null)
     {
         this.config = config;
         var _loggerFactory = loggerFactory ?? NullLoggerFactory.Instance;
         this._logger = _loggerFactory.CreateLogger<SimpleCacheClient>();
         ValidateRequestTimeout(config.TransportStrategy.GrpcConfig.DeadlineMilliseconds);
-        Claims claims;
```
kvcache, line 38: excellent, thanks

src/Momento.Sdk/Auth/EnvironmentTokenProvider.cs
```diff
@@ -0,0 +1,38 @@
...
+using System;
+using Momento.Sdk.Exceptions;
+using Momento.Sdk.Internal;
+
+public class EnvironmentTokenProvider : ICredentialProvider
+{
+
+    private string _authToken;
+    private string _controlEndpoint;
+    private string _cacheEndpoint;
```
kvcache, line 12: why not remove these and include a private set on the implementations below? You don't have to put the private set in the interface to put it on the implementation

kvcache (commented):
Couple nits, this is nice. Are you going to make other clients handle the auth tuple this way too? 🙂

src/Momento.Sdk/Auth/ICredentialProvider.cs
pgautier404: Nice, thanks.

src/Momento.Sdk/Auth/EnvironmentTokenProvider.cs
pgautier404: Thanks, fixed in fb214b5.

cprice404 (dismissed):
lgtm!  would like for at least one of @rlinehan @gautamomento to approve before we merge though.

nice job!

tests/Integration/Momento.Sdk.Tests/SimpleCacheControlTest.cs
```diff
@@ -9,24 +10,30 @@ namespace Momento.Sdk.Tests;
...
 
     [Fact]
     public void SimpleCacheClientConstructor_NullJWT_InvalidJwtException()
     {
-        Assert.Throws<InvalidArgumentException>(() => new SimpleCacheClient(Configurations.Laptop.Latest, null!, defaultTtlSeconds: 10));
+        Environment.SetEnvironmentVariable("BAD_MOMENTO_AUTH_TOKEN", null);
+        Assert.Throws<InvalidArgumentException>(
+            () => new EnvironmentTokenProvider("BAD_MOMENTO_AUTH_TOKEN")
+        );
     }
```
malandis, line 37: Is there a mechanism to setup and tear down the environment variables for the tests? My concern is when the test runner runs the tests concurrently.

The tests rely on shared state (environment variables), which are not isolated between the tests. If possible we should unset the variable after the test scope.

malandis (commented):
Just one concern about the tests

gautamomento (dismissed):
Thank you for doing this. It paves a way for our customers and us to get tokens from various sources.

tests/Integration/Momento.Sdk.Tests/SimpleCacheControlTest.cs
pgautier404: It's already unset at the end of the second test, so we're good there. I added an explicit unset in the other test case in 37e2f6f.

src/Momento.Sdk/Auth/EnvironmentTokenProvider.cs
```diff
@@ -0,0 +1,34 @@
...
+{
+    public string AuthToken { get; private set; }
+    public string ControlEndpoint {get; private set; }
+    public string CacheEndpoint { get; private set; }
+
+    /// <summary>
+    /// Reads and parses a JWT token stored as an environment variable.
+    /// <param name="name">Name of the environment variable that contains the JWT token.</param>
+    /// </summary>
+    public EnvironmentTokenProvider(string name)
```
rlinehan, line 17: My only slight concern here is around the naming. When we do introduce new tokens, users might still wan to be able to set them via an environment variable. However, they'll contain something different in the claims (won't have the control and cache endpoints), so the rest of the logic in this provider won't work.

I'm not sure we have great names for these current tokens vs. new tokens (I've sometimes called current ones "legacy" but uh... that's not a good name when we haven't released new tokens 😅) - new tokens might be called "API tokens" and "session tokens" (there will be two different kinds, neither of them will have this info in claims). Maybe something like "Momento auth token" for the current ones, so maybe this should be something like `EnvironmentMomentoAuthTokenProvider` - that's kind of a mouthful though and not sure it's actually helpful. Naming is the worst.

rlinehan: FWIW a `APITokenProvider` would probably look something like (sorry I don't really know c#)

```
public APITokenProvider(string apiToken, string endpoint)
     AuthToken = apiToken;
     ControlEndpoint = $"control.{endpoint}"
     CacheEndpoint = $"cache.{endpoint}"
```

Just in case that extra context helps.

I wonder if it would be easiest for now just to have like a `MomentoAuthTokenProvider` that doesn't do any of the environment variable reading? Idk.

rlinehan (commented):
I love this! Left one concern about naming, otherwise looks good.

src/Momento.Sdk/Auth/EnvironmentTokenProvider.cs
gautamomento, line 17: I am sorry, I know I said ship it on the previous pass. I wanted to check how would this look, say when we stop parsing region out of the auth token. 

I see two options:

1. `EnvironmentTokenProvider(string authTokenEnvName, string authTokenRegionName)`
2. Or we just do this internally by saying momento always needs you to use certain environment variables - `MOMENTO_AUTH_TOKEN` and `MOMENTO_ENDPOINT`.

Both approaches have pros and cons, but seems like we are shooting for approach #1 here?

rlinehan: I think it can't just be region name though, it needs to be endpoint, and that's not something that we've really shipped yet.

IMO we would have a separate Environment provider for different types of tokens?

pgautier404: @rlinehan Since these are specifically JWT tokens, what would you think about `EnvironmentJwtTokenProvider`? The environment variable reading is just an implementation detail of the specific class, and the interface can support the `APITokenProvider` you described just fine.

pgautier404: @gautamomento My plan was to have different provider implementations for different purposes. As long as they implement `ICredentialProvider` we'll be fine. Other providers can have all their info passed into the constructor like the `APITokenProvider` example @rlinehan posted or can read the info from files or whatever. The specific provider in this PR is here to support the example scripts and tests, which all read a JWT token from an environment variable.

rlinehan: Hmmm the new tokens will also be JWTs though, so I'm not quite sure that works. 🤔

rlinehan (approved):

PR #278, opened 2022-10-06
fix: re-work logging config

cprice404:
This commit does some re-working on how we expose and curry the
logging config (ILoggerFactory).

The motivation for this PR is that many (most?) Middlewares and
Retry Strategies will need access to the logging config, and it
is very cumbersome to try to pass it in as a constructor arg
to every single one of those.  Especially in the pre-built configs,
which currently do not even have direct access to the ILoggerFactory.

To address this we move the ILoggerFactory to be a first-class
member of the IConfiguration itself.  We also alter the constructors
for the Configuration class such that it will loop over all of the
Middlewares and the RetryStrategy and set the ILoggerFactory on them
if they do not already have one.

This ensures that customers only ever need to explictly pass the
ILoggerFactory to at most one of the constructors (the Configuration),
but we will have access to it from all of the Middlewares and other
objects that are encapsulated within the Configuration.

### DIFF
--- a/src/Momento.Sdk/Config/Configuration.cs
+++ b/src/Momento.Sdk/Config/Configuration.cs
@@ -1,5 +1,7 @@
 using System.Collections.Generic;
 using System.Linq;
+using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Logging.Abstractions;
 using Momento.Sdk.Config.Middleware;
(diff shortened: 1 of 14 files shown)


malandis (approved):
:sparkles:

pgautier404 (approved):

PR #280, opened 2022-10-06
fix: handle AggregateExceptions from task continuations

cprice404:
In Middleware code, if you need to modify the response before it
makes it back out to the caller, you need to use `Task.ContinueWith`.
However, if an exception is thrown while such a Middleware is
in the stack, the exception will get wrapped in an AggregateException.

Prior to this commit our CacheExceptionMapper would treat this as
an unknown error and we would lose our desired status code mapping.
This commit unwraps AggregateExceptions before the mapping so that
the error codes will be correct.

It also adds debug logs to the exception mapper so that we can
see what's going wrong if other exception types fail to map
properly in the future.

### DIFF
--- a/src/Momento.Sdk/Exceptions/CacheExceptionMapper.cs
+++ b/src/Momento.Sdk/Exceptions/CacheExceptionMapper.cs
@@ -1,5 +1,7 @@
 ﻿using System;
+using System.Threading.Tasks;
 using Grpc.Core;
+using Microsoft.Extensions.Logging;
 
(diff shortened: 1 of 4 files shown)


poppoerika (approved):
Looks good to me!

pgautier404 (dismissed):
Nice!

cprice404:
had to rebase this one

poppoerika (approved):

pgautier404 (approved):

Issue #223, opened 2022-09-28
re-enable warning for missing documentation & fill in missing documentation


cprice404:
We do not need to spend time on this for internal classes.  So for anything that is defined in the `Internal` directory, we can just use a compiler directive to disable warnings for those files.  TODO: @malandis add the magical incantation to the ticket.

malandis:
To disable warnings in a whole file, add this to the top of the file:

`#pragma warning disable CS1591`

Note you can also use this to disable the warning in a particular region of a file as well as follows:

```csharp

// code here; subject documentation warnings

#pragma warning disable CS1591
// more code here; documentation warnings will be suppressed
#pragma warning restore CS1591```

PR #282, opened 2022-10-06
fix: re-enable warning and address 20 warnings

poppoerika:
This pr address the first 20 warnings of 165 warning we have for documentation.
Relevant ticket: https://github.com/momentohq/client-sdk-dotnet/issues/223

### DIFF
--- a/src/Momento.Sdk/Auth/EnvMomentoTokenProvider.cs
+++ b/src/Momento.Sdk/Auth/EnvMomentoTokenProvider.cs
@@ -4,10 +4,22 @@ using System;
 using Momento.Sdk.Exceptions;
 using Momento.Sdk.Internal;
 
+/// <summary>
+/// Reads and parses a JWT token stored as an environment variable.
+/// </summary>
 public class EnvMomentoTokenProvider : ICredentialProvider
 {
+    /// <summary>
+    /// JWT provided by user
+    /// </summary>
     public string AuthToken { get; private set; }
-    public string ControlEndpoint {get; private set; }
+    /// <summary>
+    /// Control endpoint extracted from JWT
+    /// </summary>
+    public string ControlEndpoint { get; private set; }
+    /// <summary>
+    /// Cache endpoint extracted from JWT
+    /// </summary>
     public string CacheEndpoint { get; private set; }
 
     /// <summary>
@@ -16,16 +28,19 @@ public class EnvMomentoTokenProvider : ICredentialProvider
     /// <param name="name">Name of the environment variable that contains the JWT token.</param>
     public EnvMomentoTokenProvider(string name)
     {
-        AuthToken =  Environment.GetEnvironmentVariable(name);
+        AuthToken = Environment.GetEnvironmentVariable(name);
         if (String.IsNullOrEmpty(AuthToken))
         {
             throw new InvalidArgumentException($"Environment variable '{name}' is empty or null.");
         }
 
         Claims claims;
-        try {
+        try
+        {
             claims = JwtUtils.DecodeJwt(AuthToken);
-        } catch (InvalidArgumentException) {
+        }
+        catch (InvalidArgumentException)
+        {
             throw new InvalidArgumentException("The supplied Momento authToken is not valid.");
         }
         ControlEndpoint = claims.ControlEndpoint;
--- a/src/Momento.Sdk/Auth/ICredentialProvider.cs
+++ b/src/Momento.Sdk/Auth/ICredentialProvider.cs
@@ -12,11 +12,11 @@ public interface ICredentialProvider
     /// </summary>
     string AuthToken { get; }
     /// <summary>
-    /// Control endpoint extracted from JWT
+    /// Control endpoint
     /// </summary>
     string ControlEndpoint { get; }
     /// <summary>
-    /// Cache endpoint extracted from JWT
+    /// Cache endpoint
     /// </summary>
     string CacheEndpoint { get; }
 }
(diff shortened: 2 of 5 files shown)


src/Momento.Sdk/Auth/ICredentialProvider.cs
```diff
@@ -2,9 +2,21 @@ namespace Momento.Sdk.Auth;
...
+    /// </summary>
     string AuthToken { get; }
-    string ControlEndpoint {get; }
+    /// <summary>
+    /// Control endpoint extracted from JWT
+    /// </summary>
+    string ControlEndpoint { get; }
+    /// <summary>
+    /// Cache endpoint extracted from JWT
+    /// </summary>
```
pgautier404, line 20: In the future, the endpoints won't be embedded in the JWT, so these interface comments should probably be more general. I'd suggest we just use something like "Momento authentication token", "Control endpoint", and "Cache endpoint" as the summaries for these.

poppoerika: 👍 
As I was adding comments, I wasn't sure how descriptive we want those properties to be, especially these are hard to describe more than "auth token" and "endpoints" 😄

src/Momento.Sdk/Config/Configurations.cs
```diff
@@ -5,6 +5,9 @@
 
 namespace Momento.Sdk.Config;
 
+/// <summary>
+/// Provide pre-build configurations.
```
cprice404, line 9: s/build/built

src/Momento.Sdk/Auth/EnvMomentoTokenProvider.cs
```diff
@@ -4,10 +4,22 @@ namespace Momento.Sdk.Auth;
...
+    /// JWT provided by user
+    /// </summary>
     public string AuthToken { get; private set; }
-    public string ControlEndpoint {get; private set; }
+    /// <summary>
+    /// Control endpoint extracted from JWT
+    /// </summary>
+    public string ControlEndpoint { get; private set; }
+    /// <summary>
+    /// Cache endpoint extracted from JWT
```
cprice404, line 21: For these two I think you can just use inheritdoc to get the docs for whatever we put on the interface (@malandis @pgautier404  correct me if that sounds wrong.)

src/Momento.Sdk/Auth/ICredentialProvider.cs
```diff
@@ -2,9 +2,21 @@ namespace Momento.Sdk.Auth;
 
 using System.Collections.Generic;
 
+/// <summary>
+/// Interface for credentials
```
cprice404, line 6: Best practice for documentation strings:

If your docstring uses the same words as the definition/name, then it's not really adding any value.  So e.g. you don't need to say `interface` here, that will already be visible to the user.

Maybe say something like:

Specifies the fields that are required for a Momento client to connect to and authenticate with the Momento service.

pgautier404 (approved):

src/Momento.Sdk/Auth/ICredentialProvider.cs
```diff
@@ -2,9 +2,21 @@ namespace Momento.Sdk.Auth;
...
+/// </summary>
 public interface ICredentialProvider
 {
+    /// <summary>
+    /// JWT provided by user
+    /// </summary>
     string AuthToken { get; }
-    string ControlEndpoint {get; }
+    /// <summary>
+    /// Control endpoint
```
cprice404, line 15: same comment here; this doesn't give any new information beyond what the name would already provide to the user.

Here I think you might want to say something like "the host and port on which the Momento client will connect to the Momento control plane, in the format '<host>:<port>'" or something. 

(I don't know if that's the actual format, you should base64-decode that segment of your JWT and verify that before updating the docs.

src/Momento.Sdk/Config/Configuration.cs
```diff
@@ -8,20 +8,39 @@
 
 namespace Momento.Sdk.Config;
 
+/// <inheritdoc cref="Momento.Sdk.Config.IConfiguration" />
```
cprice404, line 11: 👍

src/Momento.Sdk/Config/Configuration.cs
```diff
@@ -8,20 +8,39 @@
...
 public class Configuration : IConfiguration
 {
+    /// <inheritdoc cref="Microsoft.Extensions.Logging.ILoggerFactory" />
```
cprice404, line 14: I like this, if it works the way we want it to.  @malandis is this a reasonable thing to do?  Annotating a Property with an inheritdoc for a Class?

src/Momento.Sdk/Config/Configuration.cs
```diff
@@ -33,26 +52,51 @@ public Configuration(IRetryStrategy retryStrategy, IList<IMiddleware> middleware
         this.TransportStrategy = transportStrategy;
     }
 
+    /// <summary>
+    ///  Configures logging
```
cprice404, line 56: This is technically a copy constructor so we should probably allude to that.

"Creates a new instance of the Configuration object, updated to use the specified logger factory."

src/Momento.Sdk/Config/Configuration.cs
```diff
@@ -33,26 +52,51 @@ public Configuration(IRetryStrategy retryStrategy, IList<IMiddleware> middleware
...
+    /// </summary>
+    /// <param name="loggerFactory">This is responsible for configuraing logging.</param>
+    /// <returns>Configuration object with custom logging provided</returns>
     public IConfiguration WithLoggerFactory(ILoggerFactory loggerFactory)
     {
         return new Configuration(RetryStrategy, Middlewares, TransportStrategy, loggerFactory);
     }
 
+    /// <summary>
+    ///  Configures retry strategy
```
cprice404, line 66: Same comment for the rest of these copy constructors.

cprice404 (commented):
thanks for doing this!  left a few small nits to help you level up on some best practices for writing docstrings.  Overall looks great though!

PR #283, opened 2022-10-07
feat: add basic retry middleware

cprice404:
This commit adds an initial implementation of a retry middleware, along with fleshing out the FixedCountRetryStrategy for compatibility with the new middleware.  As of this commit, any Get/Set requests that fail with an Unavailable or Internal error gRPC status code will be retried up to 3 attempts.

There is no backoff or jitter yet; we should implement that in the future.

The determination of which gRPC status codes and request types are eligible for retries is implemented via a new IRetryEligibilityStrategy interface.  This way, individual retry strategies have the ability to override the behavior, but in the 90% case they can just re-use the default eligibility strategy.

NOTE: we implement retries as a bespoke middleware rather than using the retry configuration that is built into the .NET gRPC library because the built-in implementation has some strict restrictions that aren't compatible with our goals here (e.g. no response that includes any headers is eligible for retry in the built-in implementation, and we always get headers back from envoy.)

### DIFF
--- a/src/Momento.Sdk/Config/Retry/FixedCountRetryStrategy.cs
+++ b/src/Momento.Sdk/Config/Retry/FixedCountRetryStrategy.cs
@@ -1,22 +1,31 @@
+using System.Net.NetworkInformation;
+using Grpc.Core;
 using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Logging.Abstractions;
+using Momento.Sdk.Internal.Retry;
 
 namespace Momento.Sdk.Config.Retry;
 
 public class FixedCountRetryStrategy : IRetryStrategy
 {
     public ILoggerFactory? LoggerFactory { get; }
+
+    private ILogger _logger;
+    private readonly IRetryEligibilityStrategy _eligibilityStrategy;
+
     public int MaxAttempts { get; }
 
-    //FixedCountRetryStrategy(retryableStatusCodes = DEFAULT_RETRYABLE_STATUS_CODES, maxAttempts = 3),
-    public FixedCountRetryStrategy(int maxAttempts, ILoggerFactory? loggerFactory = null)
+    public FixedCountRetryStrategy(int maxAttempts, IRetryEligibilityStrategy? eligibilityStrategy = null, ILoggerFactory? loggerFactory = null)
     {
         LoggerFactory = loggerFactory;
+        _logger = (loggerFactory ?? NullLoggerFactory.Instance).CreateLogger<FixedCountRetryStrategy>();
+        _eligibilityStrategy = eligibilityStrategy ?? new DefaultRetryEligibilityStrategy(loggerFactory);
         MaxAttempts = maxAttempts;
     }
 
     public FixedCountRetryStrategy WithLoggerFactory(ILoggerFactory loggerFactory)
     {
-        return new(MaxAttempts, loggerFactory);
+        return new(MaxAttempts, _eligibilityStrategy.WithLoggerFactory(loggerFactory), loggerFactory);
     }
 
     IRetryStrategy IRetryStrategy.WithLoggerFactory(ILoggerFactory loggerFactory)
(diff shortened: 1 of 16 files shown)


cprice404:
i am planning to add another commit that adds some unit tests, but wanted to get this up for review since I believe the bulk of the work is done.

pgautier404 (dismissed):
LGTM

cprice404:
rebased

eaddingtonwhite (approved):

src/Momento.Sdk/Config/Retry/FixedCountRetryStrategy.cs
```diff
@@ -26,15 +35,23 @@ IRetryStrategy IRetryStrategy.WithLoggerFactory(ILoggerFactory loggerFactory)
...
     public FixedCountRetryStrategy WithMaxAttempts(int maxAttempts)
     {
-        return new(maxAttempts, LoggerFactory);
+        return new(maxAttempts, _eligibilityStrategy, LoggerFactory);
     }
 
-    public int? DetermineWhenToRetryRequest(IGrpcResponse grpcResponse, IGrpcRequest grpcRequest, int attemptNumber)
+    public int? DetermineWhenToRetryRequest<TRequest>(Status grpcStatus, TRequest grpcRequest, int attemptNumber) where TRequest : class
     {
+        _logger.LogDebug($"Determining whether request is eligible for retry; status code: {grpcStatus.StatusCode}, request type: {grpcRequest.GetType()}, attemptNumber: {attemptNumber}, maxAttempts: {MaxAttempts}");
```
kvcache, line 43: Here and elsewhere I notice you're charging the users' runtimes string interpolation allocations unconditionally. I think you should rewrite all these LogDebugs like:
```suggestion
        _logger.LogDebug(
            "Determining whether request is eligible for retry; status code: {StatusCode}, request type: {RequestType}, attemptNumber: {AttemptNumber}, maxAttempts: {MaxAttempts}",
            grpcStatus.StatusCode,
            grpcRequest.GetType(),
            attemptNumber,
            MaxAttempts
        );
```

see for more examples and  detail: https://learn.microsoft.com/en-us/dotnet/api/microsoft.extensions.logging.loggerextensions.logdebug?view=dotnet-plat-ext-6.0

src/Momento.Sdk/Internal/Middleware/RetryMiddleware.cs
```diff
@@ -0,0 +1,89 @@
...
+            return WithLoggerFactory(loggerFactory);
+        }
+
+        public async Task<MiddlewareResponseState<TResponse>> WrapRequest<TRequest, TResponse>(
+            TRequest request,
+            CallOptions callOptions,
+            Func<TRequest, CallOptions, Task<MiddlewareResponseState<TResponse>>> continuation
+        ) where TRequest : class where TResponse : class
+        {
+            var foo = request.GetType();
```
kvcache, line 42: this should be removed

src/Momento.Sdk/Internal/Middleware/RetryMiddleware.cs
```diff
@@ -0,0 +1,89 @@
...
+            int? retryAfterMillis = 0;
+            do
+            {
+                var delay = retryAfterMillis ?? 0;
+                if (delay > 0)
+                {
+                    await Task.Delay(delay);
+                }
+                attemptNumber++;
+                nextState = await continuation(request, callOptions);
```
kvcache, line 54: I'm not sure I see how this will disambiguate between the likes of ListPushBack() or DictionaryIncrement() (non-idempotent methods) and Get() or Set() (idempotent methods). What's your plan there?

src/Momento.Sdk/Config/Retry/FixedCountRetryStrategy.cs
cprice404: good call, agreed, it was my intent to write them that way but I always forget.  thx

src/Momento.Sdk/Internal/Middleware/RetryMiddleware.cs
cprice404: See the code in `DefaultRetryEligibilityStrategy`, which is called as part of the `FixedCountRetryStrategy`.

src/Momento.Sdk/Config/Retry/FixedCountRetryStrategy.cs
kvcache: kotlin-structured-logging is an addictive library. It's so nice to not have to care about anything w.r.t. logging invocation cost inside of those logging blocks. I catch myself writing logs like this too so I look out for them

src/Momento.Sdk/Internal/Middleware/RetryMiddleware.cs
kvcache: I completely missed `_retryableRequestTypes`, thanks.

src/Momento.Sdk/Config/Retry/FixedCountRetryStrategy.cs
cprice404: i hope we will have a rust-structured-logging (or maybe one already exists in the rust universe)

PR #284, opened 2022-10-07
chore: update docstrings to follow best practices

poppoerika:
This pr addresses the feedback received from Chris in this pr: https://github.com/momentohq/client-sdk-dotnet/pull/282
I moved some of the docstrings to interfaces and reference those docstrings using `inheritdoc cref` in the classes inherits those interfaces.


malandis (commented):
For the documentation in general, I recommend fully documenting the public elements of interface files. Then in the implementations, you only need to write

```csharp
/// <inheritdoc />
```
above the methods/properties from the interface and it will resolve automatically, ie no need for a cref.

For example in IConfiguration.cs:

```csharp
public interface IConfiguration
{
      /// document this one
      public ILoggerFactory LoggerFactor { get; }
```

Then in Configuration.cs:

```csharp
{
     /// <inheritdoc />
     public ILoggerFactory LoggerFactory { get; }
```

As an example see the documention of ISimpleCacheClient.cs and SimpleCacheClient.cs

src/Momento.Sdk/Config/IConfiguration.cs
```diff
@@ -12,16 +12,47 @@ namespace Momento.Sdk.Config;
 /// </summary>
 public interface IConfiguration
 {
+    /// <inheritdoc cref="Momento.Sdk.Config.IConfiguration.LoggerFactory" />
     public ILoggerFactory LoggerFactory { get; }
```
pgautier404, line 16: This looks like it's trying to inherit from itself. This should (I think) reference `Microsoft.Extensions.Logging.ILoggerFactory` instead, but I'm wondering if these crefs are necessary at all? It looked like `<inheritdoc />` was working fine without that attribute, and I think @malandis was suggesting that usage in his comment (please correct me if I'm wrong about that).

pgautier404 (changes requested):

src/Momento.Sdk/Config/IConfiguration.cs
poppoerika: Thank you for pointing it out!
I moved the summaries to the interfaces and forgot to remove `cref`. I will remove them.

src/Momento.Sdk/Config/Configuration.cs
```diff
@@ -11,29 +11,24 @@ namespace Momento.Sdk.Config;
...
-    /// <param name="loggerFactory">This is responsible for configuraing logging.</param>
     public Configuration(IRetryStrategy retryStrategy, ITransportStrategy transportStrategy, ILoggerFactory? loggerFactory = null)
         : this(retryStrategy, new List<IMiddleware>(), transportStrategy, loggerFactory)
     {
 
     }
 
     /// <summary>
-    /// <inheritdoc cref="Momento.Sdk.Config.IConfiguration" />
+    /// Create a new instance of Configuration obejct with provided arguments: <see cref="Momento.Sdk.Config.IConfiguration.RetryStrategy">, <see cref="Momento.Sdk.Config.IConfiguration.Middlewares">, <see cref="Momento.Sdk.Config.IConfiguration.TransportStrategy"/>, and <see cref="Momento.Sdk.Config.IConfiguration.LoggerFactory"/>
```
pgautier404, line 31: These `see` elements need to be closed: `<see cref="foo" />`.

pgautier404, line 31: These `see` tags need to be closed: `<see cref="foo" />`

pgautier404 (approved):

malandis (dismissed):
Thank you for your diligence on this

pgautier404 (approved):

malandis (approved):

PR #286, opened 2022-10-07
feat: use `TimeSpan` to specify TTL and deadline

malandis:
This updates the code across the board to use `TimeSpan` instead of `uint` (milli)?seconds. `TimeSpan` is more ergonomic and used broadly in other client APIs. The latter type is discouraged in .NET, as it is not broadly compatible with other .NET languages. See here for more details: https://learn.microsoft.com/en-us/dotnet/standard/language-independence


src/Momento.Sdk/ISimpleCacheClient.cs
```diff
@@ -86,7 +86,7 @@ public interface ISimpleCacheClient : IDisposable
     /// <param name="cacheName">Name of the cache to store the item in.</param>
     /// <param name="key">The key to set.</param>
     /// <param name="value">The value to be stored.</param>
-    /// <param name="ttlSeconds">TTL for the item in cache. This TTL takes precedence over the TTL used when initializing a cache client. Defaults to client TTL.</param>
+    /// <param name="ttlSeconds">TTL for the item in cache. This TTL takes precedence over the TTL used when initializing a cache client. Defaults to client TTL. If specified must be strictly positive.</param>
```
kvcache, line 89: why should we regress the C# experience now instead of adding a compatibility layer for less-featured languages later?

malandis: Leaving language compatibility aside, I see this as in improvement in the experience.

While I agree the domain of values of `uint` is better suited for this than the domain of `int`, we've received feedback from two independent users that expressing TTL as an `int` is a pain because they would like to use native, ergonomic time span objects to express the TTL, eg:

```csharp
client.SetAsync("my cache", "my key", "my value", new TimeSpan(hours: 24, minutes: 0, seconds: 0).TotalSeconds);
```

but those `TimeSpan` objects deal in `int`, not `uint`, so they would have to add [checked casts](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/statements/checked-and-unchecked) everywhere, or else the compiler complains. In reality these casts are a don't care, because the range of values `uint` contains that `int` does not widely exceeds our max TTL.

We've received feedback from an expert user that `uint` is odd to see in .NET APIs.

kvcache: Huh, til my c# experience is irredeemably out of date with the style of the cool kids. Do we at least still think Jon Skeet is the bee's knees?

Maybe instead of a domain type _regression_ the _addition_ of a TimeSpan ttl option would be better? For example I would rather write
```
client.SetAsync("my cache", "my key", "my value", new TimeSpan(hours: 24, minutes: 0, seconds: 0));
```

than
```
client.SetAsync("my cache", "my key", "my value", new TimeSpan(hours: 24, minutes: 0, seconds: 0).TotalSeconds);
```

and tbh, TimeSpan might be the overwhelmingly preferred TTL type option - though that suspicion is based on an admittedly stale experience.

kvcache: also would let me write:
```
client.SetAsync("my cache", "my key", "my value", currentCacheRefreshMark - DateTime.Now);
```

to coordinate item lifetimes across my cache.

src/Momento.Sdk/ISimpleCacheClient.cs
```diff
@@ -86,7 +86,7 @@ public interface ISimpleCacheClient : IDisposable
     /// <param name="cacheName">Name of the cache to store the item in.</param>
     /// <param name="key">The key to set.</param>
     /// <param name="value">The value to be stored.</param>
-    /// <param name="ttlSeconds">TTL for the item in cache. This TTL takes precedence over the TTL used when initializing a cache client. Defaults to client TTL.</param>
+    /// <param name="ttl">TTL for the item in cache. This TTL takes precedence over the TTL used when initializing a cache client. Defaults to client TTL. If specified must be strictly positive.</param>
```
cprice404, line 89: is this comment still relevant?  Can TimeSpans be negative?

src/Momento.Sdk/Internal/ScsDataClient.cs
```diff
@@ -16,16 +14,16 @@ namespace Momento.Sdk.Internal;
...
+    protected readonly ILogger _logger;
     protected readonly CacheExceptionMapper _exceptionMapper;
 
-    public ScsDataClientBase(IConfiguration config, string authToken, string endpoint, uint defaultTtlSeconds)
+    public ScsDataClientBase(IConfiguration config, string authToken, string endpoint, TimeSpan defaultTtl)
     {
         this.grpcManager = new(config, authToken, endpoint);
-        this.defaultTtlSeconds = defaultTtlSeconds;
-        this.dataClientOperationTimeoutMilliseconds = config.TransportStrategy.GrpcConfig.DeadlineMilliseconds;
+        this.defaultTtlMilliseconds = (int)defaultTtl.TotalMilliseconds;
```
cprice404, line 25: this cast here jumps out at me.  What is the actual type of that value?  And should we be considering changing the type of our variable to match it rather than casting?

src/Momento.Sdk/Internal/ScsDataClient.cs
```diff
@@ -16,16 +14,16 @@ namespace Momento.Sdk.Internal;
 public class ScsDataClientBase : IDisposable
 {
     internal readonly DataGrpcManager grpcManager;
-    private readonly uint defaultTtlSeconds;
-    private readonly uint dataClientOperationTimeoutMilliseconds;
-    private readonly ILogger _logger;
+    protected readonly int defaultTtlMilliseconds;
+    protected readonly int dataClientOperationTimeoutMilliseconds;
+    protected readonly ILogger _logger;
```
cprice404, line 19: curious why the increase in visibility on these

cprice404 (approved):
This looks fine to me.  There is more casting than I would prefer but I'm not familiar enough with the data types to know whether any of that is avoidable, so I will defer to your judgment if you think it's necessary, and in that case, 🚢

src/Momento.Sdk/ISimpleCacheClient.cs
malandis: Yes [they can be negative](https://learn.microsoft.com/en-us/dotnet/api/system.timespan?view=net-7.0), eg they can represent the 24 hours ago relative to an index time.

src/Momento.Sdk/Internal/ScsDataClient.cs
malandis: That got clobbered in the several merges and got through in spite of looking this over several times. Thank you. Fixed in 7e1af1e.

src/Momento.Sdk/Internal/ScsDataClient.cs
malandis: Good catch. The deadline is a `DateTime`, so we do not need to convert to milliseconds (unlike before). I've changed this in  4528491. There is no longer any casting in deadline path.

malandis:
> This looks fine to me. There is more casting than I would prefer but I'm not familiar enough with the data types to know whether any of that is avoidable, so I will defer to your judgment if you think it's necessary, and in that case, ship

Thank you for considering this. There are three consumers of `TimeSpan` in the code:
- the control client deadline
- the data client deadline
- data client operations that use a TTL

The two deadlines expect a `DateTime` object. I've changed the calculations to offset from `now()` using the `TimeSpan` directly, so there is no casting there.

As for the TTL, our protos expect TTL in millis as a `ulong`. We cannot avoid casting in this case because `TimeSpan.TotalMilliseconds` returns a `double`, as fractional millis are allowed. We are safe going from `double` to `ulong` though, because we already check `TimeSpan` is positive and the absolute magnitude in millis of a `TimeSpan` < `Int64.MAX_VALUE`

src/Momento.Sdk/Internal/ScsDataClient.cs
```diff
@@ -36,17 +34,24 @@ protected Metadata MetadataWithCache(string cacheName)
...
+    /// (2) we know <paramref name="ttl.TotalMilliseconds"/> is less than <see cref="Int64.MaxValue"/>
+    /// because <see cref="TimeSpan"/> counts number of ticks, 1ms = 10,000 ticks, and max number of
+    /// ticks is <see cref="Int64.MaxValue"/>.
+    /// </remark>
+    /// <param name="ttl">The TTL to convert. Defaults to defaultTtl</param>
+    /// <returns>Milliseconds representation of the TTL (if provided, else of <see cref="defaultTtl"/>)</returns>
+    protected ulong TtlToMilliseconds(TimeSpan? ttl = null)
     {
-        return (ttlSeconds ?? defaultTtlSeconds) * 1000;
+        return (ulong)((ttl ?? defaultTtl).TotalMilliseconds);
```
cprice404, line 54: Could you use `Convert.ToUint64()` here?  In some languages a cast just means "pretend like this block of memory was this other data type".  IDK how c# represents doubles vs longs in memory or what it does with a cast like this but I would be surprised if the memory representation of the double would yield the same value if interpreted as a long without any conversion, so it seems like the `Convert` call would be safer/clearer.

src/Momento.Sdk/MomentoSigner.cs
```diff
@@ -99,11 +99,11 @@ private string CreateJwtToken(JwtPayload jwtPayload)
     }
 
 
-    private JwtPayload CommonJwtBody(string cacheName, string cacheKey, uint expiryEpochSeconds)
+    private JwtPayload CommonJwtBody(string cacheName, string cacheKey, TimeSpan expiryEpoch)
     {
         return new JwtPayload()
             {
-                { "exp", expiryEpochSeconds },
+                { "exp", (long)expiryEpoch.TotalSeconds },
```
cprice404, line 106: same comment here about convert vs. cast.

src/Momento.Sdk/Internal/ScsDataClient.cs
malandis: I'll add the statement. When jumping between languages the conventions vary so I agree it's clearer.

(Just leaving this as some documentation) .NET  does an actual conversion though, [see here](https://learn.microsoft.com/en-us/dotnet/csharp/programming-guide/types/casting-and-type-conversions#explicit-conversions). To contrast going from `uint` to `int` (where you could have a positive integer outside the range of `int`) requires [extra validation](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/statements/checked-and-unchecked).

src/Momento.Sdk/MomentoSigner.cs
malandis: Fixed in [d077b1a](https://github.com/momentohq/client-sdk-dotnet/pull/286/commits/d077b1a51c992a9a81dd8dfc0ae1dc0716301ff2)

cprice404 (dismissed):

src/Momento.Sdk/Internal/DataGrpcManager.cs
```diff
@@ -168,7 +168,7 @@ public async Task<_ListLengthResponse> ListLengthAsync(_ListLengthRequest reques
     }
 }
 
-internal class DataGrpcManager : IDisposable
+public class DataGrpcManager : IDisposable
```
cprice404, line 171: Motivation for visibility modifier change?

malandis: Incubating repo needs access to this

cprice404 (dismissed):
just the one question about the visibility modifier (which I am mostly asking about because of the fact that everything we mark as public will trigger that docstring warning :) ).

Other than that looks fine, and if those modifier changes are necessary for Incubating or some other reason, nbd.

cprice404 (approved):

PR #288, opened 2022-10-07
feat: simplify logging configuration

cprice404:
Kenny called out on a previous PR that there was some code smell in the way I was managing the logging factories for the various components of the configuration.

This commit reworks things slightly so that the user still only has to pass in the ILoggerFactory in one spot in our pre-built configs, but we now just pass that directly through to the constructors of the other things that need it.  This allows us to get rid of the getters and copy constructors that we had previously exposed.


kvcache (dismissed):
this looks great to me. I think IMiddleware is significantly better this way, and it's more clear to me now how to configure logging. The burden of invoking a static function instead of invoking a property to me seems very light and a welcome price to pay for the clarity.

malandis (dismissed):
I really like how the retry logic ended up

cprice404:
rebased

eaddingtonwhite (approved):

PR #299, opened 2022-10-11
feat: Rename `Exception` to `InnerException`

cprice404:
This commit changes the name of the `Exception` object on our
Error classes to `InnerException`.  This seems like the pattern
that is used in the existing C# exception object model for
cases like ours, where you don't usually expect the caller to
dig into the nested error details but you want to make it possible
for them to do so.

### DIFF
--- a/src/Momento.Sdk/Responses/CacheDeleteResponse.cs
+++ b/src/Momento.Sdk/Responses/CacheDeleteResponse.cs
@@ -35,8 +35,8 @@ public abstract class CacheDeleteResponse
             _error = error;
         }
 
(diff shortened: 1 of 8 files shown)


cprice404:
@malandis @pgautier404 @poppoerika lmk what you all think about this; I don't feel strongly about it, but this is how these types of exceptions seemed to be expose by, e.g., the Task framework when I was working on the retry stuff.

pgautier404:
> @malandis @pgautier404 @poppoerika lmk what you all think about this; I don't feel strongly about it, but this is how these types of exceptions seemed to be expose by, e.g., the Task framework when I was working on the retry stuff.



I think that's clearer.

poppoerika (approved):
It makes sense to me.

malandis (approved):

PR #300, opened 2022-10-11
chore: zero inbox on docstring warnings

cprice404:
This fills in the remaining docstrings to get rid all of the remaining compiler warnings.

### DIFF
--- a/src/Momento.Sdk/Config/Configurations.cs
+++ b/src/Momento.Sdk/Config/Configurations.cs
@@ -14,8 +14,8 @@ namespace Momento.Sdk.Config;
 public class Configurations
 {
     /// <summary>
-    /// Laptop config provides defaults suitable for a medium-to-high-latency dev environment.  Permissive timeouts, retries, potentially
-    /// a higher number of connections, etc.
+    /// Laptop config provides defaults suitable for a medium-to-high-latency dev environment.  Permissive timeouts, retries, and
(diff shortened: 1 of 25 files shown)


bruuuuuuuce (approved):
:shipit:

malandis (commented):
There are various empty summary and param elements. Was that intentional?

cprice404:
> There are various empty summary and param elements. Was that intentional?

yes, asterisk.

I only did that in places where I felt like the comments on the class header or the names of the parameters were already providing all the information someone would want/need.  I don't like to add docstrings that just repeat the exact same words as the variable names if they don't provide any additional useful information.

That said I'm also not super familiar yet with how/when these are made visible to the user, or what they will look like in HTML once we start publishing them.

With all of that in mind I decided to err on the side of not sinking a bunch of time into it in places where the value wasn't clear to me.  You have more experience with these than I do so if there are particular places where you believe the empty values will be problematic please let me know.

malandis:
> With all of that in mind I decided to err on the side of not sinking a bunch of time into it in places where the value wasn't clear to me. You have more experience with these than I do so if there are particular places where you believe the empty values will be problematic please let me know.

Totally get it. I just wasn't sure if those were placeholders you intended to get back to later.

malandis (dismissed):

malandis (approved):

PR #311, opened 2022-10-18
chore: rename tests to reflect errors instead of throwing exception

malandis:
The new API returns errors instead of throwing exceptions. The tests were written for exceptions, updated to test for errors, but the function names were out of date.


pgautier404 (approved):
Thanks!

PR #324, opened 2022-11-14
chore: truncate long CacheGetResponse.Hit string representations

### LINKED ISSUE #314: Revisit ToString implementation for response classes
Currently the response classes that hold a value emit that value as part of their `ToString()` override, but those values can be unsuitable for representation in that way (e.g., super long, binary, etc.). We need to settle on a sensible way to represent this value information.

malandis:
This adds extension methods to both pretty print byte arrays and truncate strings. With these extension methods we trim the string representation of `CacheGetResponse.Hit`.

Closes #314


cprice404 (dismissed):
measure

pgautier404 (approved):
LGTM!

PR #337, opened 2022-12-02
feat: remove deprecated call

malandis:
With the new version of the service, we will not isolate keys based on
type. Because of this, we do not need a dedicated `delete` RPC for
each type.

addresses https://github.com/momentohq/client-sdk-dotnet-incubating/issues/65


src/Momento.Sdk/Internal/DataGrpcManager.cs
```diff
@@ -102,12 +101,6 @@ public async Task<_DictionaryFetchResponse> DictionaryFetchAsync(_DictionaryFetc
         return await wrapped.ResponseAsync;
     }
 
-    public async Task<_DictionaryDeleteResponse> DictionaryDeleteAsync(_DictionaryDeleteRequest request, CallOptions callOptions)
```
cprice404, line 105: do we not have one of these for lists or sets?

cprice404 (approved):

src/Momento.Sdk/Internal/DataGrpcManager.cs
malandis: We do. The RPCs that delete a whole list or whole set have multiple uses, so we need to keep those. Because the dictionary delete RPC only has that one use, it's the only one we can delete outright.

PR #341, opened 2022-12-05
fix: reverts pr #337 (commit a394a)

malandis:
Previously we removed the `DictionaryDeleteAsync` method believing it
was single use for deleting an entire dictionary. To the contrary, the
RPC is used for deleting field(s) as well, so we must restore it.


cprice404 (approved):

malandis:
Previously we removed the DictionaryDeleteAsync method believing it
was single use for deleting an entire dictionary. To the contrary, the
RPC is used for deleting field(s) as well, so we must restore it.

PR #347, opened 2022-12-10
feat: retry data structure commands

malandis:
Adds support to retry idempotent data structure operations. The only
commands at this point that are not idempotent are (i) ones that add or
delete from a list positionally, and (ii) increment.

closes https://github.com/momentohq/client-sdk-dotnet-incubating/issues/55

### DIFF
--- a/src/Momento.Sdk/Internal/Retry/DefaultRetryEligibilityStrategy.cs
+++ b/src/Momento.Sdk/Internal/Retry/DefaultRetryEligibilityStrategy.cs
@@ -55,6 +55,12 @@ namespace Momento.Sdk.Internal.Retry
             // not idempotent: typeof(_ListPopFrontRequest),
             // not idempotent: typeof(_ListPopBackRequest),
             typeof(_ListFetchRequest),
+            /*
+             *  Warning: in the future, this may not be idempotent
+             *  Currently it supports removing all occurrences of a value.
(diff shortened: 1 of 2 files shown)


src/Momento.Sdk/Internal/Retry/DefaultRetryEligibilityStrategy.cs
```diff
@@ -40,7 +40,31 @@ public class DefaultRetryEligibilityStrategy : IRetryEligibilityStrategy
...
+            // not idempotent: typeof(_ListPopFrontRequest),
+            // not idempotent: typeof(_ListPopBackRequest),
+            typeof(_ListFetchRequest),
+            /*
+             *  Warning: in the future, this may not be idempotent
+             *  Currently it supports removing all occurrences of a value.
+             *  In the future, we may also add "the first/last N occurrences of a value".
+             *  In the latter case it is not idempotent.
+             */
+            typeof(_ListRemoveRequest),
```
malandis, line 64: Unlike other commands which are either idempotent or not, this one (in the future) will be idempotent in some cases and not in others.

This is a case where a command can be idempotent depending on a field. If we choose to remove all elements equal to a value, then the command is idempotent. But if we choose to remove, say, the first (last) element equal to a value, then it may not be.

Currently we only support removing all elements equal to a value, hence I have left it retryable.

cprice404 (approved):

PR #348, opened 2022-12-14
feat: add support for configuring eager connection on client

cprice404:
This commit adds a configuration setting that can be used to enable eager connection to the server when the client is constructed.  By default, the client establishes the connection lazily when the first request is executed.  However, this approach can lead to request timeouts on the first request if the connection takes too long to establish, because the connection time is counted toward the client-side timeout. When eager connections are enabled, the client will attempt to establish the connection to the server at construction time, which means that the connection will already be established before the first user request is issued.

### DIFF
--- a/src/Momento.Sdk/Auth/EnvMomentoTokenProvider.cs
+++ b/src/Momento.Sdk/Auth/EnvMomentoTokenProvider.cs
@@ -9,6 +9,8 @@ using Momento.Sdk.Internal;
 /// </summary>
 public class EnvMomentoTokenProvider : ICredentialProvider
 {
+    private readonly string envVarName;
+    
     /// <inheritdoc />
     public string AuthToken { get; private set; }
     /// <inheritdoc />
@@ -22,6 +24,8 @@ public class EnvMomentoTokenProvider : ICredentialProvider
     /// <param name="name">Name of the environment variable that contains the JWT token.</param>
     public EnvMomentoTokenProvider(string name)
     {
+        this.envVarName = name;
+        
         AuthToken = Environment.GetEnvironmentVariable(name);
         if (String.IsNullOrEmpty(AuthToken))
         {
@@ -31,5 +35,12 @@ public class EnvMomentoTokenProvider : ICredentialProvider
         var claims = AuthUtils.TryDecodeAuthToken(AuthToken);
         ControlEndpoint = claims.ControlEndpoint;
         CacheEndpoint = claims.CacheEndpoint;
+    }
+
+    public ICredentialProvider WithCacheEndpoint(string cacheEndpoint)
+    {
+        var updated = new EnvMomentoTokenProvider(this.envVarName);
+        updated.CacheEndpoint = cacheEndpoint;
+        return updated;
     }
 }
--- a/src/Momento.Sdk/Auth/EnvMomentoTokenProvider.cs
+++ b/src/Momento.Sdk/Auth/EnvMomentoTokenProvider.cs
@@ -37,6 +37,7 @@ public class EnvMomentoTokenProvider : ICredentialProvider
         CacheEndpoint = claims.CacheEndpoint;
     }
 
+    /// <inheritdoc />
     public ICredentialProvider WithCacheEndpoint(string cacheEndpoint)
     {
         var updated = new EnvMomentoTokenProvider(this.envVarName);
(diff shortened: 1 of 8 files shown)


gautamomento (commented):
Thank you @cprice404 , this looks good to me. 

I have left one question.

kvcache (commented):
Looks good to me once gautam is happy

src/Momento.Sdk/Internal/DataGrpcManager.cs
```diff
@@ -224,7 +225,23 @@ internal DataGrpcManager(IConfiguration config, string authToken, string host)
             }
         ).ToList();
 
-        Client = new DataClientWithMiddleware(new Scs.ScsClient(invoker), middlewares);
+        var client = new Scs.ScsClient(invoker);
+
+        if (config.TransportStrategy.EagerConnection)
+        {
+            _logger.LogDebug("TransportStrategy EagerConnection is enabled; attempting to connect to server");
+            var pingClient = new Ping.PingClient(this.channel);
```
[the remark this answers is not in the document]
cprice404: this is a fair callout.  I added a commit that changes the setting from a boolean to a timespan, and then sets a deadline based on the timespan.

tests/Integration/Momento.Sdk.Tests/SimpleCacheEagerConnectionTest.cs
```diff
@@ -0,0 +1,50 @@
...
+        var client = new SimpleCacheClient(config, authProvider, defaultTtl);
+    }
+    
+    
+    [Fact]
+    public void SimpleCacheClientConstructor_EagerConnection_BadEndpoint()
+    {
+        var config = Configurations.Laptop.Latest(loggerFactory);
+        config = config.WithTransportStrategy(config.TransportStrategy.WithEagerConnectionTimeout(TimeSpan.FromSeconds(2)));
+        var authProviderWithBadCacheEndpoint = authProvider.WithCacheEndpoint("cache.cell-external-beta-1.prod.a.momentohq.com:65000");
```
cprice404, line 46: This host/port combination will hang indefinitely, so the test hits the timeout.

gautamomento (dismissed):
LGTM

src/Momento.Sdk/Internal/DataGrpcManager.cs
```diff
@@ -224,7 +225,25 @@ internal DataGrpcManager(IConfiguration config, string authToken, string host)
...
+        {
+            TimeSpan eagerConnectionTimeout = config.TransportStrategy.EagerConnectionTimeout.Value;
+            _logger.LogDebug("TransportStrategy EagerConnection is enabled; attempting to connect to server");
+            var pingClient = new Ping.PingClient(this.channel);
+            try
+            {
+                pingClient.Ping(new _PingRequest(),
+                    new CallOptions(deadline: DateTime.UtcNow.Add(eagerConnectionTimeout)));
+            }
+            catch (RpcException ex)
```
malandis, line 240:
```suggestion
            catch (RpcException)
```

src/Momento.Sdk/Auth/EnvMomentoTokenProvider.cs
```diff
@@ -32,4 +36,11 @@ public EnvMomentoTokenProvider(string name)
         ControlEndpoint = claims.ControlEndpoint;
         CacheEndpoint = claims.CacheEndpoint;
     }
+
+    public ICredentialProvider WithCacheEndpoint(string cacheEndpoint)
```
malandis, line 40: missing documentation here

malandis (commented):
Looks great. Small nits

src/Momento.Sdk/Auth/EnvMomentoTokenProvider.cs
cprice404: good catch, will fix before moving this out of draft

cprice404:
Update: I verified that the server-side changes related to this have been deployed to all the cells.  I've rebased this and believe it's ready to go in now.

src/Momento.Sdk/Auth/EnvMomentoTokenProvider.cs
cprice404: fixed in https://github.com/momentohq/client-sdk-dotnet/pull/348/commits/56c8d2b7de59913bae02b97897b210049e2f5bda

kvcache (approved):
This looks good to me, thanks

malandis (approved):

PR #354, opened 2023-01-06
feat: update load-gen example

### LINKED ISSUE #229: implement ellery's suggestions about using time-based constructs for loadgen, etc.
Make the load-gen example match the Python and Javascript examples.

- [x] logLevel
- [x] printStatsEveryNRequests -> showStatsInterval
  - [x] Takes a TimeSpan.
- [x]  totalNumberOfOperationsToExecute -> howLongToRun (?)
  - [x] Takes a TimeSpan.
- [x] Display maxRequestsPerSecond, numberOfConcurrentRequests, and howLongToRun to the user before running.

I skipped requestTimeout, the example uses the pre-built Laptop config.

These are very desirable UX improvements because with the current setup, if the user changes `maxRequestsPerSecond` or `numberOfConcurrentRequests`, it will have a big impact on how often (in wall clock time) the stats are printed, as well as how long the program will run for.  That's not intuitive for users.

schwern:
This is my first time doing dotnet; be savage.

Closes #229

### DIFF
--- a/examples/MomentoLoadGen/Program.cs
+++ b/examples/MomentoLoadGen/Program.cs
@@ -399,7 +399,7 @@ If you have questions or need help experimenting further, please reach out to us
               ///
               logLevel: LogLevel.Debug,
               ///
-              /// Each time this amount of time has past, statistics about throughput and latency
+              /// Each time this amount of time has passed, statistics about throughput and latency
               /// will be printed.
               /// 
               ///
--- a/examples/MomentoLoadGen/Program.cs
+++ b/examples/MomentoLoadGen/Program.cs
@@ -335,7 +335,6 @@ cumulative get latencies:
 
         private static string OutputHistogramSummary(HistogramBase histogram)
         {
-            Console.WriteLine($"Histogram totalcount: {histogram.TotalCount}");
             return $@"
 count: {histogram.TotalCount}
         p50: {histogram.GetValueAtPercentile(50)}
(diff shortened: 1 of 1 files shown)


examples/MomentoLoadGen/Program.cs
```diff
@@ -407,45 +393,45 @@ static ILoggerFactory InitializeLogging()
...
+            CsharpLoadGeneratorOptions loadGeneratorOptions = new CsharpLoadGeneratorOptions(
+              ///
+              /// Controls the verbosity of the output during the run.
+              ///
+              logLevel: LogLevel.Debug,
+              ///
+              /// Each time the load generator has executed this many requests, it will
+              /// print out some statistics about throughput and latency.
+              ///
+              showStatsInterval: new TimeSpan(0, 0, 5),
```
cprice404, line 405: one small nit - I believe the TimeSpan class has some static methods that are factory constructors that are something like "ofSeconds" or "ofMinutes".  I find those to be easier to read/review.

cprice404 (changes requested):
this looks great other than one very small nit

examples/MomentoLoadGen/Program.cs
schwern: And I forgot to change the comment. 🙄

examples/MomentoLoadGen/Program.cs
```diff
@@ -399,10 +399,11 @@ static async Task Main(string[] args)
               ///
               logLevel: LogLevel.Debug,
               ///
-              /// Each time the load generator has executed this many requests, it will
-              /// print out some statistics about throughput and latency.
+              /// Each time this amount of time has past, statistics about throughput and latency
```
cprice404, line 402: nit: s/past/passed?

cprice404, line 402:
```suggestion
              /// Each time this amount of time has passed, statistics about throughput and latency
```

cprice404 (approved):
:ty:

PR #363, opened 2023-02-13
chore: update examples to 1.4.0, add example for eager connection config

cprice404:
We have a few customers that need to use the eager connection
configuration to deal with cold start issues.  This commit adds
an example of how to use this configuration.

### DIFF
--- a/examples/MomentoApplication/Program.cs
+++ b/examples/MomentoApplication/Program.cs
@@ -17,7 +17,7 @@ using (SimpleCacheClient client = new SimpleCacheClient(Configurations.Laptop.La
     await AdvancedExamples.ListCachesExample(client);
     await AdvancedExamples.SetGetDeleteExample(client);
     await AdvancedExamples.DeleteCacheExample(client);
-    await AdvancedExamples.EagerConnectionExample();
+    AdvancedExamples.EagerConnectionExample();
 
 
     Console.WriteLine("\nProgram has completed successfully.");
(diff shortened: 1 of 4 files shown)


examples/MomentoApplication/Program.cs
```diff
@@ -8,97 +8,139 @@
...
-    Console.WriteLine($"\nDeleting cache {CACHE_NAME}");
-    var deleteCacheResponse = await client.DeleteCacheAsync(CACHE_NAME);
-    if (deleteCacheResponse is DeleteCacheResponse.Error deleteCacheError)
+    /// <summary>
+    /// By default, cache clients will connect lazily when the first request is
+    /// issued.  This example shows how to configure a client to connect eagerly
+    /// when it is instantiated, and to specify a distinct timeout for the
+    /// connection to be established.
+    /// </summary>
+    public static async Task EagerConnectionExample()
```
malandis, line 135: The GitHub checks complain at this line saying there are no awaited expressions below, so the function should not be async. Does the constructor await the eager connection call?

nand4011: Is there a pre-completed task that can be returned like completedFuture in Java?

cprice404: the constructor call results in a synchronous call to force the connection, so you're right, I don't need async here.

nand4011 (approved):

examples/MomentoApplication/Program.cs
cprice404: addressed in https://github.com/momentohq/client-sdk-dotnet/pull/363/commits/299c9ad80feb332b48219f4d4afc350ffc436bf9

malandis (approved):

PR #367, opened 2023-02-13
feat: support configuring a minimum number of gRPC channels

cprice404:
This commit adds a setting to gRPC config that allows users to configure a minimum number of gRPC channels.  Prior to this, the client would always construct a single channel and would dynamically add additional ones only after we exceeded the maximum number of concurrent requests on the current channels. This results in imbalanced load on the server for cases where the number of concurrent requests remains below the threshold.

Users with this type of workload can increase this new setting to ensure they have more than one connection to the servers, and thus increase the likelihood of balancing requests across multiple servers for improved performance.

### DIFF
--- a/src/Momento.Sdk/SimpleCacheClient.cs
+++ b/src/Momento.Sdk/SimpleCacheClient.cs
@@ -1,4 +1,6 @@
 ﻿using System;
+using System.Collections.Generic;
+using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.Extensions.Logging;
 using Momento.Sdk.Auth;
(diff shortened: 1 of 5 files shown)


src/Momento.Sdk/SimpleCacheClient.cs
```diff
@@ -190,14 +198,23 @@ public async Task<CacheSetResponse> SetAsync(string cacheName, string key, byte[
...
+            dataClient.Dispose();
+        }
         GC.SuppressFinalize(this);
     }
+
+
+    private ScsDataClient NextDataClient()
+    {
+        return this.dataClients[Interlocked.Increment(ref this.nextDataClientIndex) % this.dataClients.Count];
+    }
```
kvcache, line 219: Yeah this makes sense to me.

kvcache (approved):
Looks good, thanks

nand4011 (approved):

PR #385, opened 2023-02-15
feat: remove next token param from list caches

### LINKED ISSUE #382: remove next token from list caches arguments

malandis:
Pending server side support, we remove the next token param from the
list caches control operation.

Closes #382

### DIFF
--- a/src/Momento.Sdk/ISimpleCacheClient.cs
+++ b/src/Momento.Sdk/ISimpleCacheClient.cs
@@ -86,7 +86,6 @@ public interface ISimpleCacheClient : IDisposable
     /// <summary>
     /// List all caches.
     /// </summary>
-    /// <param name="nextPageToken">A token to specify where to start paginating. This is the NextToken from a previous response.</param>
     /// <returns>
     /// Task representing the result of the list cache operation. The
     /// response object is resolved to a type-safe object of one of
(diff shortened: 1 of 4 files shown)


cprice404:
> Note: though we remove the argument from the API, we leave next token in the response class.

I actually took this out in node.js in my current PR.  thoughts?

nand4011 (commented):
Are we leaving the token in the response object for backwards compatibility? The intention in the node library is to hide the token entirely from the user and handle the paging automatically once it is implemented.

malandis:
> > Note: though we remove the argument from the API, we leave next token in the response class.
> 
> I actually took this out in node.js in my current PR. thoughts?

No strong thoughts. Perhaps an unused field is more confusing than just removing it. I was going to keep it for backwards compatibility since this part of the API was already 1.0

cprice404:
> > > Note: though we remove the argument from the API, we leave next token in the response class.
> > 
> > 
> > I actually took this out in node.js in my current PR. thoughts?
> 
> No strong thoughts. Perhaps an unused field is more confusing than just removing it. I was going to keep it for backwards compatibility since this part of the API was already 1.0

we're already changing this signature though.  I think let's just remove it for now.

malandis:
Removed in https://github.com/momentohq/client-sdk-dotnet/pull/385/commits/a7ef870505ecd955861f2126c4683ef6feaf23ba

nand4011 (approved):

cprice404 (approved):

PR #389, opened 2023-02-16
fix: Improve accessibility of field name in DictionaryGetField response

cprice404:
This commit moves the definition of the `field` property up to
the parent class, DictionaryGetFieldResponse.  Prior to this
commit it was defined on each individual subclass, meaning you
could only access it within a type guard conditional.  This way,
you can access it regardless of whether your response was a Hit,
Miss, or Error.  This allows us to simplify some things in the
Dictionary example.


nand4011 (dismissed):

cprice404:
Rebased to deal with merge conflicts from @malandis 's docs PR

nand4011 (approved):

malandis (approved):

PR #401, opened 2023-03-09
feat: rename SimpleCacheClient to CacheClient

cprice404:
This commit renames the primary client object from SimpleCacheClient
to CacheClient, after we received feedback from users that the
name was misleading and seemed to imply that there were multiple
clients to choose from.

The legacy name is retained as a type alias with a deprecation
warning so that this change is backward-compatible.


malandis:
> The legacy name is retained as a type alias with a deprecation warning so that this change is backward-compatible.

I can't seem to find this. Can you point me to the file/line number?

cprice404:
> > The legacy name is retained as a type alias with a deprecation warning so that this change is backward-compatible.
> 
> I can't seem to find this. Can you point me to the file/line number?

The diff is horrific but it's here: https://github.com/momentohq/client-sdk-dotnet/pull/401/files#diff-d0d1e6e67925f89694a318f907ec4e23d32f0a6b8c28466c62fbe4f67c29bbdcR13

cprice404:
I think I screwed something up in this branch, plus the diff is impossible to read, so I think I'm just going to do this over again quickly.

PR #402, opened 2023-03-09
chore: update examples to use CacheClient instead of SimpleCacheClient

cprice404:
For early review.  Need to get rid of the csproj changes once we have a release that contains the new name.

### DIFF
--- a/examples/DictionaryExample/DictionaryExample.csproj
+++ b/examples/DictionaryExample/DictionaryExample.csproj
@@ -9,9 +9,6 @@
 
   <ItemGroup>
     <PackageReference Include="Microsoft.Extensions.Logging.Console" Version="6.0.0" />
-    <PackageReference Include="Momento.Sdk" Version="1.6.0" />
-  </ItemGroup>
-  <ItemGroup>
-    <None Remove="Momento.Sdk" />
+    <PackageReference Include="Momento.Sdk" Version="1.8.0" />
   </ItemGroup>
 </Project>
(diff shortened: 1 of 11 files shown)


malandis (dismissed):

tests/Unit/Momento.Sdk.Tests/CacheClientConstructorTest.cs
```diff
@@ -0,0 +1,25 @@
+﻿using System;
+using Momento.Sdk;
+using Momento.Sdk.Auth;
+using Momento.Sdk.Config;
+using Xunit;
+
+public class CacheClientConstructorTest
+{
+    const String TEST_AUTH_TOKEN = "eyJhbGciOiJIUzUxMiJ9.eyJzdWIiOiJmb29Abm90LmEuZG9tYWluIiwiY3AiOiJjb250cm9sLXBsYW5lLWVuZHBvaW50Lm5vdC5hLmRvbWFpbiIsImMiOiJjYWNoZS1lbmRwb2ludC5ub3QuYS5kb21haW4ifQo.rtxfu4miBHQ1uptWJ2x3UiAwwJYcMeYIkkpXxUno_wIavg4h6YJStcbxk32NDBbmJkJS7mUw6MsvJNWaxfdPOw";
```
allenheltondev, line 9: Shouldn't this be an environment variable or some other substituted value so we don't check an auth token into source control?

allenheltondev (changes requested):
What is the convention for using test auth tokens? I don't think we should check one into source control.

malandis (approved):

PR #405, opened 2023-03-09
fix: added byte key and string value; fixed the order of key and value

nidhidham:


src/Momento.Sdk/ISimpleCacheClient.cs
```diff
@@ -195,6 +195,9 @@ public interface ISimpleCacheClient : IDisposable
     /// <inheritdoc cref="SetAsync(string, byte[], byte[], TimeSpan?)"/>
     public Task<CacheSetResponse> SetAsync(string cacheName, string key, byte[] value, TimeSpan? ttl = null);
 
+    /// <inheritdoc cref="SetAsync(string, byte[], byte[], TimeSpan?)"/>
+    public Task<CacheSetResponse> SetAsync(string cacheName, byte[] key, string value, TimeSpan? ttl = null);
```
cprice404, line 199: nice find!

Can I recommend two small cleanups while you are in here?

1. Can you re-organize these so that all the Sets are together, all the Gets are together, etc.?  It seems confusing and error prone for us to have spread them out this way (not your fault!)  If that seems like a ton of work it's okay to punt it for now.
2. Can you see what happens if you change this line:
https://github.com/momentohq/client-sdk-dotnet/blob/main/tests/Integration/Momento.Sdk.Tests/Fixtures.cs#L15

to use the Interface instead of the class?  I think that would catch if there are any other places where we've missed stuff in the interface?

src/Momento.Sdk/Internal/ScsDataClient.cs
```diff
@@ -78,7 +78,7 @@ public ScsDataClient(IConfiguration config, string authToken, string endpoint, T
 
     public async Task<CacheSetResponse> SetAsync(string cacheName, byte[] key, byte[] value, TimeSpan? ttl = null)
     {
-        return await this.SendSetAsync(cacheName, value: value.ToByteString(), key: key.ToByteString(), ttl: ttl);
+        return await this.SendSetAsync(cacheName, key: key.ToByteString(), value: value.ToByteString(),  ttl: ttl);
```
cprice404, line 81: thanks for changing this!  it's definitely more consistent and readable this way.

nidhidham: Not sure how this was not caught in the test cases. The test cases use Utils.NewGuidByteArray() to generate keys and values, so ideally, it should have been captured there!

cprice404: for posterity: we discussed this on a zoom - it's because of the keyword args here.  even though they were passed in a weird order, it worked, because of the keyword args.

malandis:
I had actually removed byte key and string value from a previous version. 😬  Do we need this in the public API? I had expected to add the various permutations of {string, byte[]} key and {string, byte[]} value pending demand.

src/Momento.Sdk/ISimpleCacheClient.cs
nidhidham: Thank you! 
1. Fixed the re-organization of sets/gets/deletes
2. Noob alert: If we use the interface, not sure how we can set the Logger - https://github.com/momentohq/client-sdk-dotnet/blob/main/tests/Integration/Momento.Sdk.Tests/Fixtures.cs#L26

nidhidham:
> I had actually removed byte key and string value from a previous version. 😬 Do we need this in the public API? I had expected to add the various permutations of {string, byte[]} key and {string, byte[]} value pending demand.

Not sure if we need it 👀

malandis:
> Not sure if we need it 👀

I see these combinations are in the Java SDK (have been there for a while). I'm not clear we definitely needed there either. I'll let @cprice404 weigh in. Regardless I appreciate the eye for detail and completeness 🙇‍♂️

nidhidham:
> > Not sure if we need it 👀
> 
> I see these combinations are in the Java SDK (have been there for a while). I'm not clear we definitely needed there either. I'll let @cprice404 weigh in. Regardless I appreciate the eye for detail and initiative 🙇‍♂️

Thank you :)

src/Momento.Sdk/ISimpleCacheClient.cs
cprice404: Awesome, thanks!

Re: 2. -> you will need to change that constructor call from `new(...)` to `new CacheClient(...)`.  The `new(...)` syntax is some fancy C# syntax sugar that allows you to omit the class name if the compiler can infer it.  You are correct that when you change the return type to the interface, that fancy syntax sugar will no longer work because the compiler can no longer infer the type that you are trying to construct.  But you can just explicitly specify it rather than using the syntax sugar.  Good question!

cprice404:
> I had actually removed byte key and string value from a previous version. 😬 Do we need this in the public API? I had expected to add the various combinations of {string, byte[]} key and {string, byte[]} value pending demand.

oh wow, that makes a ton of sense, I had forgotten that we had done that intentionally.

I agree with Michael, it's probably better to leave them out until someone asks for them; fewer things to maintain.  I like the rest of the changes you made in this PR though, especially if you are able to change the test fixture to use the interface, so I would still like to keep the rest of the PR!

Sorry I didn't catch this sooner, and I agree with Michael - awesome initiative and eye for detail so thank you!

cprice404:
cc: @nand4011 ^^ there are comments on this PR that are relevant to your forthcoming Java work

nidhidham:
> oh wow, that makes a ton of sense, I had forgotten that we had done that intentionally.
> 
> I agree with Michael, it's probably better to leave them out until someone asks for them; fewer things to maintain. I like the rest of the changes you made in this PR though, especially if you are able to change the test fixture to use the interface, so I would still like to keep the rest of the PR!
> 
> Sorry I didn't catch this sooner, and I agree with Michael - awesome initiative and eye for detail so thank you!

Thank you!!! Removed {key: byte[], value: string} and updated `Fixture.cs`!

cprice404 (approved):
:ty:

PR #406, opened 2023-03-10
feat: Change name of SimpleCacheClient to CacheClient

cprice404:
As per feedback from dev advocates and users, the name SimpleCacheClient
is a little confusing and we prefer CacheClient.  This commit
implements the name change.

NOTE: we need to retain a deprecated alias using the old name,
so that we don't break backward compatibility.  That will be done
in a separate commit because the diff is too complex to review
when it is done in the same commit.


tests/Integration/Momento.Sdk.Tests/CacheEagerConnectionTest.cs
```diff
@@ -21,30 +21,30 @@ public class SimpleCacheEagerConnectionTest
...
+    public void CacheClientConstructor_EagerConnection_BadEndpoint()
     {
         var config = Configurations.Laptop.Latest(loggerFactory);
         config = config.WithTransportStrategy(config.TransportStrategy.WithEagerConnectionTimeout(TimeSpan.FromSeconds(2)));
         var authProviderWithBadCacheEndpoint = authProvider.WithCacheEndpoint("cache.cell-external-beta-1.prod.a.momentohq.com:65000");
         // validating that the constructor doesn't fail when the eager connection fails
-        var client = new SimpleCacheClient(config, authProviderWithBadCacheEndpoint, defaultTtl);
+        var client = new CacheClient(config, authProviderWithBadCacheEndpoint, defaultTtl);
     }
 }
```
malandis, line 50: Dropped a newline here

malandis (commented):
Looks good; just a small formatting tweak

tests/Integration/Momento.Sdk.Tests/CacheEagerConnectionTest.cs
cprice404, line 50:
```suggestion
}

```

cprice404: fixed in https://github.com/momentohq/client-sdk-dotnet/pull/406/commits/4857da7e78172a73f148fa8ccadb2872a3df8765

src/Momento.Sdk/Momento.Sdk.csproj
```diff
@@ -57,13 +57,13 @@
...
 	<ProjectExtensions>
-		<MonoDevelop>
-			<Properties>
-				<Policies>
-					<TextStylePolicy TabWidth="4" IndentWidth="4" RemoveTrailingWhitespace="True" NoTabsAfterNonTabs="False" EolMarker="Native" FileWidth="80" TabsToSpaces="True" scope="application/xml" />
-					<XmlFormattingPolicy scope="application/xml" />
-				</Policies>
-			</Properties>
-		</MonoDevelop>
+	  <MonoDevelop>
```
allenheltondev, line 60: These are the kind of updates I like to see!!

allenheltondev (approved):

PR #409, opened 2023-03-15
feat: add support for KeyExists and KeysExist

cprice404:


src/Momento.Sdk/Responses/CacheKeysExistResponse.cs
```diff
@@ -0,0 +1,118 @@
...
+        /// request exists in the cache or not.  The order is the same as the order that
+        /// the keys were specified in.
+        /// </summary>
+        public IEnumerable<bool> ExistsEnumerable { get; private set; }
+
+        /// <summary>
+        /// A dictionary whose keys are the cache keys from the request, and whose values are
+        ///  True if the specified key exists in the cache, false otherwise.
+        /// </summary>
+        public IDictionary<string, bool> ExistsDictionary
```
poppoerika, line 53: This is not something we have in the Go SDK.
I'd assume we had a specific .NTE customer request for this?
Even so, should we make it availabe in the Go SDK as well?

tests/Integration/Momento.Sdk.Tests/CacheDataTest.cs
```diff
@@ -39,7 +196,7 @@ public async Task SetAsync_InvalidTTLByteArrayByteArray_IsError(int ttlSeconds)
         Assert.Equal(MomentoErrorCode.INVALID_ARGUMENT_ERROR, errorResponse.ErrorCode);
     }
 
-    // Tests SetAsyc(cacheName, byte[], byte[]) as well as GetAsync(cacheName, byte[])
+    // Tests SetAsync(cacheName, byte[], byte[]) as well as GetAsync(cacheName, byte[])
```
poppoerika, line 199: 🦅 👁️

tests/Integration/Momento.Sdk.Tests/CacheDataTest.cs
```diff
@@ -11,6 +14,160 @@ public CacheDataTest(CacheClientFixture fixture) : base(fixture)
...
+            [key3] = true
+        };
+        Assert.Equal(expectedDict, goodResponse.ExistsDictionary);
+
+        var keysInADifferentOrder = new string[] { key2, key3, key1 };
+        response = await client.KeysExistAsync(cacheName, keysInADifferentOrder);
+        Assert.True(response is CacheKeysExistResponse.Success, $"Unexpected response: {response}");
+        goodResponse = (CacheKeysExistResponse.Success)response;
+        Assert.Equal(new List<bool> { false, true, true }, goodResponse.ExistsEnumerable.ToList());
+        Assert.Equal(expectedDict, goodResponse.ExistsDictionary);
```
poppoerika, line 167: I felt it's a bit confusing that the order of `ExistsEnumerable` is the same order as the passed keys, `{ key2, key3, key1 }`, but the order of `ExistsDictionary` is different, `{
            [key1] = true,
            [key2] = false,
            [key3] = true
        }`.
        
Is this expected? Am I missing something here?

src/Momento.Sdk/Responses/CacheKeysExistResponse.cs
cprice404: We can do it on-demand, or whenever we have time.  I was just trying to think of how I would be most likely to want to use this if I was writing code against it, and most likely the first thing I would do with an "array of booleans" type of response is dump it into a Map/Dictionary so that I can then write a conditional around `resultsDict["foo"]` rather than having to correlate my keys with the booleans based on their index.  So it seems like a really convenient way to access the data.

I'm open to considering omitting it but with my "user" hat on, I think this is what I would want.

tests/Integration/Momento.Sdk.Tests/CacheDataTest.cs
cprice404: The basic Dictionary/Map/Associative Array classes in almost all programming languages are un-ordered.  So the equality check is just based on whether the key/value pairs all match.

cprice404:
@malandis I've done a bit more research and I am feeling okay about the API names here.  Accessor names/types are up for discussion so if you have any feedback I'd love to hear it.

tests/Integration/Momento.Sdk.Tests/CacheDataTest.cs
poppoerika: Got it.

tests/Integration/Momento.Sdk.Tests/CacheDataTest.cs
```diff
@@ -11,6 +14,160 @@ public CacheDataTest(CacheClientFixture fixture) : base(fixture)
...
+        Assert.True(response is CacheKeysExistResponse.Error, $"Unexpected response: {response}");
+        Assert.Equal(MomentoErrorCode.INVALID_ARGUMENT_ERROR, ((CacheKeysExistResponse.Error)response).ErrorCode);
+    }
+
+    [Fact]
+    public async Task KeysExistAsync_ByteArray()
+    {
+        byte[] key1 = Encoding.UTF8.GetBytes("foo");
+        byte[] key2 = Encoding.UTF8.GetBytes("bar");
+        byte[] key3 = Encoding.UTF8.GetBytes("baz");
```
malandis, line 120:
```suggestion
        byte[] key1 = Utils.NewGuidByteArray();
        byte[] key2 = Utils.NewGuidByteArray();
        byte[] key3 = Utils.NewGuidByteArray();
```

tests/Integration/Momento.Sdk.Tests/CacheDataTest.cs
```diff
@@ -11,6 +14,160 @@ public CacheDataTest(CacheClientFixture fixture) : base(fixture)
...
+        goodResponse = (CacheKeysExistResponse.Success)response;
+        Assert.Equal(new List<bool> { false, true, true }, goodResponse.ExistsEnumerable.ToList());
+    }
+
+    [Fact]
+    public async Task KeysExistAsync_String()
+    {
+        var key1 = "foo";
+        var key2 = "bar";
+        var key3 = "baz";
```
malandis, line 143:
```suggestion
        var key1 = Utils.NewGuidString();
        var key2 = Utils.NewGuidString();
        var key3 = Utils.NewGuidString();
```

src/Momento.Sdk/Responses/CacheKeysExistResponse.cs
```diff
@@ -0,0 +1,118 @@
...
+            }
+        }
+
+        /// <include file="../docs.xml" path='docs/class[@name="Success"]/description/*' />
+        public Success(IEnumerable<ByteString> keys, _KeysExistResponse response)
+        {
+            this.keys = keys.ToList();
+            ExistsEnumerable = response.Exists;
+
+        }
```
malandis, line 74: I recommend the Lazy pattern here

https://github.com/momentohq/client-sdk-dotnet/blob/main/src/Momento.Sdk/Responses/CacheListFetchResponse.cs#L49-L79

```suggestion
        protected readonly Lazy<IDictionary<string, bool>> _existsDictionary;
        public IDictionary<string, bool> ExistsDictionary { get => _existsDictionary.Value; }

        /// <include file="../docs.xml" path='docs/class[@name="Success"]/description/*' />
        public Success(IEnumerable<ByteString> keys, _KeysExistResponse response)
        {
            this.keys = keys.ToList();
            ExistsEnumerable = response.Exists;
            _existsDictionary = new(() =>
            {
                Dictionary<string, bool> result = new();

                var i = 0;
                foreach (var exists in ExistsEnumerable)
                {
                    result[keys[i].ToStringUtf8()] = exists;
                    i++;
                }
                return result;
            });
        }
```

malandis (commented):
Few small changes

src/Momento.Sdk/Responses/CacheKeysExistResponse.cs
```diff
@@ -0,0 +1,119 @@
...
+        /// the keys were specified in.
+        /// </summary>
+        public IEnumerable<bool> ExistsEnumerable { get; private set; }
+
+        /// <summary>
+        /// A dictionary whose keys are the cache keys from the request, and whose values are
+        ///  True if the specified key exists in the cache, false otherwise.
+        /// </summary>
+        protected readonly Lazy<IDictionary<string, bool>> _existsDictionary;
+        public IDictionary<string, bool> ExistsDictionary { get => _existsDictionary.Value; }
```
malandis, line 56:
```suggestion
        protected readonly Lazy<IDictionary<string, bool>> _existsDictionary;
        /// <summary>
        /// A dictionary whose keys are the cache keys from the request, and whose values are
        ///  True if the specified key exists in the cache, false otherwise.
        /// </summary>
        public IDictionary<string, bool> ExistsDictionary { get => _existsDictionary.Value; }
```

malandis (commented):
Documentation needs to be on the public member. Other than that looks good to go

PR #415, opened 2023-03-17
chore: add `startIndex` and `endIndex` to `listFetch` API

### LINKED ISSUE #410: add start/stop index to listFetch

poppoerika:
- Added additional params, `startIndex` and `endIndex` to the `listFetch` API
- Added tests for the new params

Closes #410

### DIFF
--- a/src/Momento.Sdk/CacheClient.cs
+++ b/src/Momento.Sdk/CacheClient.cs
@@ -1060,7 +1060,7 @@ public class CacheClient : ICacheClient
     }
 
     /// <inheritdoc />
-    public async Task<CacheListFetchResponse> ListFetchAsync(string cacheName, string listName, int? startIndex, int? endIndex)
+    public async Task<CacheListFetchResponse> ListFetchAsync(string cacheName, string listName, int? startIndex = null, int? endIndex = null)
     {
         try
         {
--- a/src/Momento.Sdk/CacheClient.cs
+++ b/src/Momento.Sdk/CacheClient.cs
@@ -1060,19 +1060,24 @@ public class CacheClient : ICacheClient
     }
 
     /// <inheritdoc />
-    public async Task<CacheListFetchResponse> ListFetchAsync(string cacheName, string listName)
+    public async Task<CacheListFetchResponse> ListFetchAsync(string cacheName, string listName, int? startIndex, int? endIndex)
     {
         try
         {
             Utils.ArgumentNotNull(cacheName, nameof(cacheName));
             Utils.ArgumentNotNull(listName, nameof(listName));
+            Utils.ValidateStartEndIndex(startIndex, endIndex);
         }
         catch (ArgumentNullException e)
         {
             return new CacheListFetchResponse.Error(new InvalidArgumentException(e.Message));
         }
+        catch (Momento.Sdk.Exceptions.InvalidArgumentException e)
+        {
+            return new CacheListFetchResponse.Error(new InvalidArgumentException(e.Message));
+        }
 
-        return await this.DataClient.ListFetchAsync(cacheName, listName);
+        return await this.DataClient.ListFetchAsync(cacheName, listName, startIndex, endIndex);
     }
 
     /// <inheritdoc />
--- a/src/Momento.Sdk/Internal/Utils.cs
+++ b/src/Momento.Sdk/Internal/Utils.cs
@@ -144,14 +144,8 @@ public static class Utils
             {
                 throw new Momento.Sdk.Exceptions.InvalidArgumentException("endIndex (exclusive) must be larger than startIndex (inclusive)");
             }
-            else
-            {
-                return;
-            }
-        }
-        else
-        {
             return;
         }
+        return;
     }
 }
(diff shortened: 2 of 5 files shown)


tests/Integration/Momento.Sdk.Tests/ListTest.cs
```diff
@@ -26,6 +26,94 @@ public async Task ListConcatenateFrontAsync_NullChecksStringArray_IsError(string
...
+        CacheListFetchResponse fetchResponse = await client.ListFetchAsync(cacheName, listName, -2, -1);
+        Assert.True(fetchResponse is CacheListFetchResponse.Hit, $"Unexpected response: {fetchResponse}");
+        var hitResponse = (CacheListFetchResponse.Hit)fetchResponse;
+        Assert.Equal(new string[] { value3 }, hitResponse.ValueListString);
+    }
+
+    [Theory]
+    [InlineData("cache", "my-list", null, 1)]
+    [InlineData("cache", "my-list", null, -3)]
+    public async Task ListFetchAsync_WithNullStartIndex_HappyPath(string cacheName, string listName, int startIndex, int endIndex)
```
malandis, line 83:
```suggestion
    public async Task ListFetchAsync_WithNullStartIndex_HappyPath(string cacheName, string listName, int? startIndex, int? endIndex)
```

pgautier404 (approved):
LGTM!

src/Momento.Sdk/CacheClient.cs
```diff
@@ -1060,19 +1060,24 @@ public async Task<CacheListPopBackResponse> ListPopBackAsync(string cacheName, s
     }
 
     /// <inheritdoc />
-    public async Task<CacheListFetchResponse> ListFetchAsync(string cacheName, string listName)
+    public async Task<CacheListFetchResponse> ListFetchAsync(string cacheName, string listName, int? startIndex, int? endIndex)
```
cprice404, line 1063: so here you are specifying these as nullable, but you are not providing default values.  In the interface, you are providing default values.  i'm a little surprised that works?  Does C# implicitly default nullable values to `null`?

I would think that without the `= null`s here in this signature that you wouldn't be able to call this method without explicitly passing the nulls in, which would be a breaking change... but it looks like the old tests that were calling this w/o these arguments are still passing, so I'm missing something somewhere :)

src/Momento.Sdk/Internal/Utils.cs
```diff
@@ -125,4 +125,26 @@ public int GetHashCode(T obj)
...
     /// </summary>
     public static StructuralEqualityComparer<byte[]> ByteArrayComparer = new();
+
+    /// <summary>
+    /// Thrown an exception if the end index is larger than the start index.
+    /// </summary>
+    /// <param name="startIndex">Starting inclusive index of operation.</param>
+    /// <param name="endIndex">Ending exclusive index of operation.</param>
+    /// <exception cref="Momento.Sdk.Exceptions.InvalidArgumentException"><paramref name="endIndex"/> is less than or equal to <paramref name="startIndex"/></exception>
+    public static void ValidateStartEndIndex(int? startIndex, int? endIndex)
```
cprice404, line 135: what happens if they are null?  I don't know how C# evaluates `null >= 0`?

cprice404: i almost wonder if you only want to call this if they're both not null?  in which case you don't need the signature to allow them to be nullable?

src/Momento.Sdk/CacheClient.cs
malandis: My vote would be to make the default value explicit `= null` for both.

src/Momento.Sdk/Internal/Utils.cs
malandis: Can test explicitly with `.HasValue` as described [here](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/builtin-types/nullable-value-types)

```csharp
if (startIndex.HasValue && endIndex.HasValue)
{
```

src/Momento.Sdk/CacheClient.cs
poppoerika: I missed passing `null` there... so thank you for the call out!
I think it still works because I pass `null` in `ICacheClient` [here](https://github.com/momentohq/client-sdk-dotnet/pull/415/files#diff-9fc4e55974840be85a83f62e4c6a9e34615858796de63857408137cd172d5bc2R596)?

src/Momento.Sdk/Internal/Utils.cs
poppoerika: 👍

src/Momento.Sdk/Internal/Utils.cs
```diff
@@ -125,4 +125,33 @@ public int GetHashCode(T obj)
...
+        {
+            if (startIndex >= 0 && endIndex >= 0 && startIndex >= endIndex)
+            {
+                throw new Momento.Sdk.Exceptions.InvalidArgumentException("endIndex (exclusive) must be larger than startIndex (inclusive)");
+            }
+            if (startIndex < 0 && endIndex < 0 && startIndex >= endIndex)
+            {
+                throw new Momento.Sdk.Exceptions.InvalidArgumentException("endIndex (exclusive) must be larger than startIndex (inclusive)");
+            }
+            else
```
cprice404, line 147: supernit: I don't think that either this else clause or the one below it are technically necessary

cprice404 (approved):
one non-blocking supernit from me but feel free to ship as-is

cprice404 (approved):

PR #417, opened 2023-03-17
chore: Add lambda cdk example

rishtigupta:
## PR Description:
This commit adds 
1. A basic .net lambda which prints hello world. 
2. CDK stack that deploys the dotnet lambda to AWS Lambda

## Run this example
1.  `cd examples/LambdaExample/LambdaExampleHandler/src/LambdaExampleHandler` 
    - run `dotnet build` 
    - run `dotnet lambda package`
2. `cd examples`
    - run `dotnet build src`
    - run `AWS_PROFILE=dev cdk deploy`

## Issue 
https://github.com/momentohq/client-sdk-dotnet/issues/407

## Steps to create a lambda and deploy using cdk
TODO - notion wiki link


examples/LambdaExample/.gitignore
```diff
@@ -0,0 +1,342 @@
+# CDK asset staging directory
+.cdk.staging
+cdk.out
+
```
cprice404, line 4: holy cow this is a huge .gitignore :)

it's not a huge deal, this is probably fine, but I think I might empty this out and then just add in the stuff that you actually need as you see what shows up in the git status list.  Probably the vast majority of this we don't need and it might be confusing to people, or hard to reason about.

cprice404 (approved):
looks great!  maybe trim down the .gitignore but otherwise 🚢

cprice404 (approved):

PR #421, opened 2023-04-10
chore: update JP ver. of README

poppoerika:


pgautier404 (commented):
It looks like this and the main templated README are both still referring to `SimpleCacheClient` instead of `CacheClient`. Would you mind fixing that in this file at least and maybe opening an issue to take a closer look at the main README as well?

pgautier404 (approved):
🙇‍♂️

PR #426, opened 2023-04-26
feat: support .NET Framework 4.61+ using gRPC-Web

malandis:
Because .NET Framework does not support gRPC over HTTP/2, we fall back
to using gRPC for those older versions. To do this we:

1. Add another target framework for the library, `net461`. This is the
lowest .NET Framework version also supported by `netstandard2.0`.

2. Conditional on a `net461` build, link the gRPC-Web client
dependency.

3. Conditional on a `net461` build, use a gRPC-Web `HttpHandler` for
the control and data gRPC channels.

4. Update the test projects to exercise the `net461` build as well.

closes: https://github.com/momentohq/dev-eco-issue-tracker/issues/317

### DIFF
--- a/src/Momento.Sdk/Internal/ControlGrpcManager.cs
+++ b/src/Momento.Sdk/Internal/ControlGrpcManager.cs
@@ -1,9 +1,4 @@
-﻿#if NETFRAMEWORK
-// Because .NET Framework does not support HTTP/2, we need to use gRPC-Web over HTTP/1.1
-#define USE_GRPC_WEB
-#endif
-
-#pragma warning disable 1591
+﻿#pragma warning disable 1591
 using System;
 using System.Collections.Generic;
 using System.Threading.Tasks;
--- a/src/Momento.Sdk/Internal/ControlGrpcManager.cs
+++ b/src/Momento.Sdk/Internal/ControlGrpcManager.cs
@@ -1,9 +1,18 @@
-﻿#pragma warning disable 1591
+﻿#if NETFRAMEWORK
+// Because .NET Framework does not support HTTP/2, we need to use gRPC-Web over HTTP/1.1
+#define USE_GRPC_WEB
+#endif
+
+#pragma warning disable 1591
 using System;
 using System.Collections.Generic;
 using System.Threading.Tasks;
 using Grpc.Core;
 using Grpc.Net.Client;
+#if USE_GRPC_WEB
+using System.Net.Http;
+using Grpc.Net.Client.Web;
+#endif
 using Microsoft.Extensions.Logging;
 using Momento.Protos.ControlClient;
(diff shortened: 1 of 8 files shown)


malandis:
Opening for review. I will build a test app and manually run on Microsoft Server 2016.

src/Momento.Sdk/Internal/ControlGrpcManager.cs
```diff
@@ -1,9 +1,18 @@
-﻿#pragma warning disable 1591
+﻿#if NETFRAMEWORK
```
cprice404, line 1: under what conditions does this check succeed?  Would it be worth adding a comment here that describes the conditions?

src/Momento.Sdk/Momento.Sdk.csproj
```diff
@@ -2,7 +2,7 @@
 <Project Sdk="Microsoft.NET.Sdk">
 	<PropertyGroup>
 		<!-- Build Configuration -->
-		<TargetFramework>netstandard2.0</TargetFramework>
+		<TargetFrameworks>netstandard2.0;net461</TargetFrameworks>
```
cprice404, line 5: it's okay if we don't know the answer to this until we play with the released version a bit, but i'm curious which of these will be chosen as the framework when building for .NET Framework 4.8.x

cprice404 (commented):
this is awesome!  thanks for figuring it out.  I just had a few minor questions.

src/Momento.Sdk/Momento.Sdk.csproj
malandis: It uses `net461`. I had tested this already with the integration tests. That is, I can set the integration tests to run on .NET Framework 4.8. It will select the binary built on `net461`. See https://learn.microsoft.com/en-us/dotnet/standard/library-guidance/cross-platform-targeting#multi-targeting

cprice404: lovely

cprice404: i think it would be worth adding a comment/link on the `#if NETFRAMEWORK` lines to make it super clear when that condition is satisfied, but not a blocker.

cprice404 (dismissed):
amazing

src/Momento.Sdk/Momento.Sdk.csproj
malandis: I refactored this part in this commit. That conditional is no longer in the code and is only in the csproj file now. See https://github.com/momentohq/client-sdk-dotnet/pull/426/commits/e6276718d20e34ed5899f56a0dbef8541ce4a872

cprice404 (approved):

PR #427, opened 2023-04-26
feat: add lambda configuration

malandis:
Adds a configuration optimized for Lambda environments. This is the
same as the `InRegion.Default` config but also eagerly connects the
client. This avoids the cold start penalty from before where we
established the connection on the first request, leading to
first-request timeouts.

### DIFF
--- a/src/Momento.Sdk/Config/Configurations.cs
+++ b/src/Momento.Sdk/Config/Configurations.cs
@@ -165,8 +165,8 @@ public class Configurations
         }
 
         /// <summary>
-        /// This config optimizes for lambda environments. In addition to the low latency settings of
-        /// <see cref="LowLatency"/>, this configures the clients to eagerly connect to the Momento service
+        /// This config optimizes for lambda environments. In addition to the in region settings of
+        /// <see cref="Default"/>, this configures the clients to eagerly connect to the Momento service
         /// to avoid the cold start penalty of establishing a connection on the first request.
         /// </summary>
         public class Lambda : Configuration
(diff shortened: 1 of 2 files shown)


src/Momento.Sdk/Config/Configurations.cs
```diff
@@ -163,5 +163,33 @@ public static IConfiguration V1(ILoggerFactory? loggerFactory = null)
...
+            }
+
+            /// <summary>
+            /// Provides the latest recommended configuration for a lambda environment.
+            /// </summary>
+            /// <param name="loggerFactory"></param>
+            /// <returns></returns>
+            public static IConfiguration Latest(ILoggerFactory? loggerFactory = null)
+            {
+                var config = LowLatency.V1(loggerFactory);
```
cprice404, line 187: I think maybe you should go with `InRegion.Default` rather than `InRegion.LowLatency`.  I don't think lambdas are guaranteed to have great networking traits so making the timeout a tiny bit more lenient is probably the way to go.

Sorry for not mentioning sooner.

@eaddingtonwhite can you weigh in on this?

malandis: I changed this to `InRegion.Default` [here](https://github.com/momentohq/client-sdk-dotnet/pull/427/commits/b164b6a621a1420c8c183d93545883a3c341af16)

eaddingtonwhite: Agree w/ this. Start more conservative w/ default timeouts we can always make more agressive down line.

malandis: OK. In that case https://github.com/momentohq/client-sdk-dotnet/pull/427/commits/b164b6a621a1420c8c183d93545883a3c341af16 implements this.

cprice404 (approved):

PR #428, opened 2023-04-26
chore: print a warning before eager connection failure test

cprice404:
We have a test that asserts that the CacheClient can still be
constructed even if it's in EagerConnection mode and the Ping
fails.  That logs a scary looking warning during the int test
runs.  This commit adds a println to let future developers
know that they don't need to be worried by it :)


malandis (approved):

kvcache (commented):
image

This is really nice, thanks

PR #434, opened 2023-05-02
chore: add interface for all error response types

### LINKED ISSUE #433: Error types should implement a common interface
The current implementation treats all error types (`CacheGetResponse.Error`, `CacheSetResponse.Error`, ...) as independent from one another, even though they expose common properties such as `message`, `innerException`, and `errorCode`.

As such it is difficult to write functions against a generic Error object. To do this we have to write a function against a tuple of `(message, innerException, errorCode)` instead, which is not ergonomic.

Instead have error types implement a common error interface. That way downstream code can write code against the interface.

malandis:
In the current formulation, downstream code cannot consume a generic
`Error` type, because current error types inherit from a parent
abstract class.

This PR introduces an error interface, `IError`, which specifies that
an error response has the following properties:
- `InnerException`
- `ErrorCode`
- `Message`

which is common to all error classes.

With this in place we inherit the documentation from the interface,
simplifying the doc strings in the response classes.

Eventually we should the same for `Success`, `Hit`, and `Miss`.

closes #433

### DIFF
--- a/src/Momento.Sdk/Responses/CacheDeleteResponse.cs
+++ b/src/Momento.Sdk/Responses/CacheDeleteResponse.cs
@@ -33,7 +33,7 @@ public abstract class CacheDeleteResponse
     public class Success : CacheDeleteResponse { }
 
     /// <include file="../docs.xml" path='docs/class[@name="Error"]/description/*' />
-    public class Error : CacheDeleteResponse
+    public class Error : CacheDeleteResponse, IError
     {
         private readonly SdkException _error;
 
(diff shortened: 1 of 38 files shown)


src/Momento.Sdk/Responses/CacheDeleteResponse.cs
```diff
@@ -43,19 +43,19 @@ public Error(SdkException error)
             _error = error;
         }
 
-        /// <include file="../docs.xml" path='docs/class[@name="Error"]/prop[@name="InnerException"]/*' />
+        /// <inheritdoc />
```
eaddingtonwhite, line 46: doc inheritance 👍  😎

eaddingtonwhite (approved):

cprice404 (approved):
surprised we didn't have this already, thank you

PR #441, opened 2023-05-12
fix: update int test ttls to be more lenient

bruuuuuuuce:
When running these integration tests in region in an aws account, we were seeing flaky results. It seems to be due to the fact that we return keys/values even slightly after they are ttl'd out. This pr relaxes some of the integration test timeouts to allow some leniency in their timeouts/ttls.

Testing in fargate, these changes are passing now

### DIFF
--- a/tests/Integration/Momento.Sdk.Tests/Utils.cs
+++ b/tests/Integration/Momento.Sdk.Tests/Utils.cs
@@ -11,4 +11,12 @@ public static class Utils
     public static string NewGuidString() => Guid.NewGuid().ToString();
 
     public static byte[] NewGuidByteArray() => Guid.NewGuid().ToByteArray();
+
+    public static int initialRefreshTtl = 4;
+
+    public static int updatedRefreshTtl = 10;
+
+    public static int waitForItemToBeSet = 100;
+
+    public static int waitForInitialItemToExpire = 4900;
 }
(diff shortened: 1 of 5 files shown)


malandis (commented):
While you're here, could you extract the TTLs and wait times on these tests to static fields? A lot of the constants are repeated (mostly due to me 😬 ).

tests/Integration/Momento.Sdk.Tests/Utils.cs
```diff
@@ -11,4 +11,12 @@ public static class Utils
...
 
     public static byte[] NewGuidByteArray() => Guid.NewGuid().ToByteArray();
+
+    public static int initialRefreshTtl = 4;
+
+    public static int updatedRefreshTtl = 10;
+
+    public static int waitForItemToBeSet = 100;
+
+    public static int waitForInitialItemToExpire = 4900;
```
malandis, line 21: Style nit: use properties here. I recommend doing the rename first then adding the property getter.

```suggestion
    public static readonly int InitialRefreshTtl { get; } = 4;

    public static readonly int UpdatedRefreshTtl { get; } = 10;

    public static readonly int WaitForItemToBeSet { get; } = 100;

    public static readonly int WaitForInitialItemToExpire { get; } = 4900;

```

malandis (commented):
Thanks. This is much better. Just one last change then we're good to go.

tests/Integration/Momento.Sdk.Tests/Utils.cs
bruuuuuuuce: 👍 , apparently just adding `{ get; }` makes the entire property readonly, and if you try and set it it like `Utils.InitialRefreshTtl = 1111` its a compile time error.

malandis (approved):
❇️

PR #449, opened 2023-06-06
feat: Add ListLength to list modification apis

### LINKED ISSUE #414: add listLength to return value in all list modification operations
In this PR, we added the updated list length to all of the list modification operations:

https://github.com/momentohq/client-protos/pull/138

However, we haven't added the length to the response object in the SDKs yet.  This ticket is just for adding that field to the response objects and updating tests to cover them.

rishtigupta:
## PR Description:
- Add ListLength to list modification apis
- Reformat integration tests to cover ListLength methods

## Notes:
Adding checked after casting the `response.Found.Length` (which return a uint) to int because: uint in public APIs is discouraged in .NET as it is not CLS compliant

## Testing:
`make test-net6` to run the entire suite

## Issue
closes #414

### DIFF
--- a/src/Momento.Sdk/Responses/CacheListFetchResponse.cs
+++ b/src/Momento.Sdk/Responses/CacheListFetchResponse.cs
@@ -51,6 +51,11 @@ public abstract class CacheListFetchResponse
 #pragma warning restore 1591
 
         /// <summary>
+        /// The length of the list.
+        /// </summary>
+        public int ListLength { get; private set; }
+
+        /// <summary>
         /// 
         /// </summary>
         /// <param name="response">The cache response.</param>
@@ -66,6 +71,8 @@ public abstract class CacheListFetchResponse
             {
                 return new List<string>(values.Select(v => v.ToStringUtf8()));
             });
+
+            ListLength = checked((int)values.Count);
         }
 
         /// <summary>
@@ -83,7 +90,7 @@ public abstract class CacheListFetchResponse
         {
             var stringRepresentation = String.Join(", ", ValueListString.Select(value => $"\"{value}\""));
             var byteArrayRepresentation = String.Join(", ", ValueListByteArray.Select(value => $"\"{value.ToPrettyHexString()}\""));
-            return $"{base.ToString()}: ValueListString: [{stringRepresentation.Truncate()}] ValueListByteArray: [{byteArrayRepresentation.Truncate()}]";
+            return $"{base.ToString()}: ValueListString: [{stringRepresentation.Truncate()}] ValueListByteArray: [{byteArrayRepresentation.Truncate()}] ListLength: {ListLength}";
         }
     }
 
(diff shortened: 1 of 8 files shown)


src/Momento.Sdk/Responses/CacheListFetchResponse.cs
```diff
@@ -66,6 +71,8 @@ public Hit(_ListFetchResponse response)
             {
                 return new List<string>(values.Select(v => v.ToStringUtf8()));
             });
+
+            ListLength = checked((int)values.Count);
```
cprice404, line 75: what does `checked` do here? And what is the type of `values.Count`?

rishtigupta: I feel checked is not required here since we would not overflow the integral-limit in this case. But because the other Apis were using checked for ListLength, I was not sure if the check is applicable here or not. I can remove that. 

values.Count should return int by default. My bad, will correct that too.

src/Momento.Sdk/Responses/CacheListPopFrontResponse.cs
```diff
@@ -44,13 +44,19 @@ public class Hit : CacheListPopFrontResponse
...
+        public int ListLength { get; private set; }
+
         /// <summary>
         /// 
         /// </summary>
         /// <param name="response">The cache response</param>
         public Hit(_ListPopFrontResponse response)
         {
             this.value = response.Found.Front;
+            ListLength = checked((int)response.Found.ListLength);
```
cprice404, line 59: same question

src/Momento.Sdk/Responses/CacheListRetainResponse.cs
```diff
@@ -32,6 +33,25 @@ public abstract class CacheListRetainResponse
...
+        /// </summary>
+        public int ListLength { get; private set; }
+
+        /// <summary>
+        /// 
+        /// </summary>
+        /// <param name="response">The cache response</param>
+        public Success(_ListRetainResponse response)
+        {
+            ListLength = checked((int)response.Found.ListLength);
```
cprice404, line 47: and here

cprice404 (commented):
just a couple more occurrences where I'm not sure we need to be going the `checked` route, but I've reviewed the rest of the PR and it looks good to me!

src/Momento.Sdk/Responses/CacheListPopBackResponse.cs
```diff
@@ -44,13 +44,19 @@ public class Hit : CacheListPopBackResponse
...
+        public int ListLength { get; private set; }
+
         /// <summary>
         /// 
         /// </summary>
         /// <param name="response">The cache response.</param>
         public Hit(_ListPopBackResponse response)
         {
             this.value = response.Found.Back;
+            ListLength = checked((int)response.Found.ListLength);
```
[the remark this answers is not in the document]
rishtigupta: type of "response.Found.ListLength" is uint

cprice404 (approved):

malandis (commented):
Looks good just one questions about the scope of the change

PR #450, opened 2023-06-07
feat: Add Momento/Redis Lambda example

rishtigupta:


examples/MomentoRedisLambdaExample/MomentoRedisExampleLambdaHandler/src/MomentoRedisExampleLambdaHandler/Function.cs
```diff
@@ -0,0 +1,300 @@
...
+public record struct SetResult(CacheSetResponse Response, long ElapsedMillis);
+
+public record struct RedisGetResult(RedisValue Response, long ElapsedMillis, long ElapsedMillisWithDecompression);
+
+public record struct RedisSetResult(bool Response, long ElapsedMillis);
+
+public class Function
+{
+    const string CacheName = "MomentoLambda";
+    const string RedisClusterEndpoint = "momentoredisexamplecluster.exmof5.ng.0001.usw2.cache.amazonaws.com:6379";
```
malandis, line 39: Should extract this to a parameter

examples/MomentoRedisLambdaExample/src/MomentoRedisLambdaExample/Program.cs
```diff
@@ -0,0 +1,24 @@
...
+    {
+        public static void Main(string[] args)
+        {
+            var app = new App();
+            new MomentoRedisLambdaExampleStack(app, "MomentoRedisLambdaExampleStack", new StackProps
+            {
+                Env = new Amazon.CDK.Environment 
+                {
+                    Account = "287427698164",
+                    Region = "us-west-2"
```
malandis, line 18: Should extract these to parameters

examples/MomentoRedisLambdaExample/cdk.context.json
```diff
@@ -0,0 +1,46 @@
+{
+  "vpc-provider:account=287427698164:filter.vpc-id=vpc-00e94e9613dde8d21:region=us-west-2:returnAsymmetricSubnets=true": {
+    "vpcId": "vpc-00e94e9613dde8d21",
+    "vpcCidrBlock": "172.31.0.0/16",
+    "ownerAccountId": "287427698164",
```
malandis, line 5: Should the account IDs be in the public CDK files?

malandis (commented):
Few questions about the hardcoded constants

examples/MomentoRedisLambdaExample/MomentoRedisExampleLambdaHandler/src/MomentoRedisExampleLambdaHandler/Function.cs
```diff
@@ -0,0 +1,300 @@
...
+
+namespace MomentoRedisExampleLambdaHandler;
+
+public class Input
+{
+    public bool RunMomentoWithCompression { get; set; }
+    public bool RunMomentoCompressionWithSleep { get; set; } 
+    public bool RunMomentoWithoutCompression { get; set; }
+    public bool RunRedisWithCompression { get; set; }
+    public bool RunRedisWithoutCompression { get; set; }
```
pgautier404, line 19: Can these be narrowed down to eliminate the ones that are opposite one another? You should be able to assume that `RunMomentoWithCompression == false` is equivalent to `RunMomentoWithoutCompression` and avoid some of the input validation below. `RunMomentoWithCompressionWithSleep` could also be shortened to `CompressWithSleep` since it can just be ignored if `RunMomentoWithCompression` is false.

examples/MomentoRedisLambdaExample/MomentoRedisExampleLambdaHandler/src/MomentoRedisExampleLambdaHandler/Function.cs
```diff
@@ -0,0 +1,300 @@
...
+    }
+
+    private async Task<string> DoStuff(Input input) {
+
+        Console.WriteLine("Start executing DoStuff....");
+
+        // Create Momento client
+        ICredentialProvider authProvider = new EnvMomentoTokenProvider("MOMENTO_AUTH_TOKEN");
+        var momentoClient = new CacheClient(
+            Configurations.InRegion.Lambda.Latest().WithClientTimeout(TimeSpan.FromSeconds(3000)),
```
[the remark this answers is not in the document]
rishtigupta: yes, this is intentional. Needed so that the keys do not expire in between running tests

examples/MomentoRedisLambdaExample/MomentoRedisExampleLambdaHandler/src/MomentoRedisExampleLambdaHandler/Function.cs
rishtigupta: Will extract these as parameters to lambda function

examples/MomentoRedisLambdaExample/cdk.context.json
rishtigupta: Yes, I thought the same. But, according to their [documentation](https://docs.aws.amazon.com/cdk/v2/guide/context.html#:~:text=Because%20they%27re%20part%20of%20your%20application%27s%20state%2C%20cdk.json%20and%20cdk.context.json%20must%20be%20committed%20to%20source%20control%20along%20with%20the%20rest%20of%20your%20app%27s%20source%20code.%20Otherwise%2C%20deployments%20in%20other%20environments%20(for%20example%2C%20a%20CI%20pipeline)%20might%20produce%20inconsistent%20results.), they suggest to commit cdk.context.json. 

Do you have any suggestions on this?

cprice404: for this use case let's not commit that file, let's put it in gitignore please

examples/MomentoRedisLambdaExample/src/MomentoRedisLambdaExample/Program.cs
rishtigupta: yes, we can get these values from "Aws.ACCOUNT_ID" and "Aws.REGION".

examples/MomentoRedisLambdaExample/MomentoRedisExampleLambdaHandler/src/MomentoRedisExampleLambdaHandler/Function.cs
```diff
@@ -0,0 +1,153 @@
...
+
+        List<Task<SetResult>> momentoSetTasks;
+        List<Task<GetResult>> momentoGetTasks;
+        List<Task<RedisSetResult>> redisSetTasks;
+        List<Task<RedisGetResult>> redisGetTasks;
+        
+        momentoSetTasks = input.RunMomento ? await momentoCache.DoMomentoSets(_numMomentoSetRequests, withSleep: input.SleepMomento) : new List<Task<SetResult>>();
+        redisSetTasks = input.RunRedis ? await redisCache.DoRedisSets(_numRedisSetRequests) : new List<Task<RedisSetResult>>();
+        momentoGetTasks = input.RunMomento ? await momentoCache.DoMomentoGets(_numMomentoGetRequests, compressValue: input.CompressMomento, withSleep: input.SleepMomento) : new List<Task<GetResult>>();
+        redisGetTasks = input.RunRedis ? await redisCache.DoRedisGets(_numRedisGetRequests, compressValue: input.CompressRedis) : new List<Task<RedisGetResult>>();
```
cprice404, line 111: having these awaits here looks a little suspicious. I think that the way you've implemented it now it is probably working fine but in terms of understanding this code I think we would not expect these functions to be async or to require awaiting.  More comments a little later in the PR

examples/MomentoRedisLambdaExample/MomentoRedisExampleLambdaHandler/src/MomentoRedisExampleLambdaHandler/MomentoCache.cs
```diff
@@ -0,0 +1,188 @@
...
+        }
+    }
+    
+    public async Task<List<Task<SetResult>>> DoMomentoSets(int numMomentoSetRequests, bool withSleep = false)
+    {
+        Console.WriteLine($"\t WITH SLEEP flag: {withSleep}");
+        try
+        {
+            List<Task<SetResult>> momentoSetTasks = new List<Task<SetResult>>();
+            string sourceData = await Utils.GetSourceData();
```
cprice404, line 42: this looks to me like the only place where you are using an await in this function, meaning that if you didn't have to do this here, this function could be synchronous, which would make the calling code a little easier to understand.

What I would do is make this sourceData variable be an argument to the function. Then, you can acquire it once at the beginning of the main function (including the await that you need to do in order to realize it), and pass the string reference into this function.

examples/MomentoRedisLambdaExample/MomentoRedisExampleLambdaHandler/src/MomentoRedisExampleLambdaHandler/MomentoCache.cs
```diff
@@ -0,0 +1,188 @@
...
+            Console.WriteLine($"SET returned from sleeping 500ms for key {key}");
+        }
+        
+        stopWatch.Stop();
+        var duration = stopWatch.ElapsedMilliseconds;
+        Console.WriteLine($"\tReturning set response for {key}: {setResponse}");        
+        return new SetResult(Response: setResponse, ElapsedMillis: duration);
+    }
+    
+#pragma warning disable CS1998
```
cprice404, line 84: just curious what the warning is here.

examples/MomentoRedisLambdaExample/MomentoRedisExampleLambdaHandler/src/MomentoRedisExampleLambdaHandler/MomentoCache.cs
```diff
@@ -0,0 +1,188 @@
...
+
+        for (int i = 1; i <= numMomentoGetRequests; i++)
+        {
+            Console.WriteLine("Executing MomentoGets");
+            string key = $"key-{i}";
+            momentoGetTasks.Add(ExecuteGet(key, compressValue, withSleep));
+        }
+
+        return momentoGetTasks;
+    }
```
cprice404, line 97: there are no awaits in this function so it doesn't need to be async. maybe that's the warning you disabled above :)

Making this not be async will make the calling code more clear as per my first comment.

examples/MomentoRedisLambdaExample/MomentoRedisExampleLambdaHandler/src/MomentoRedisExampleLambdaHandler/Function.cs
```diff
@@ -0,0 +1,153 @@
...
+
+        Console.WriteLine("Awaiting pending tasks");
+        // empty lists should no-op
+        Task.WaitAll(momentoSetTasks.Select(task => (Task)task).ToArray());
+        Task.WaitAll(redisSetTasks.Select(task => (Task)task).ToArray());
+        Task.WaitAll(momentoGetTasks.Select(task => (Task)task).ToArray());
+        Task.WaitAll(redisGetTasks.Select(task => (Task)task).ToArray());
```
cprice404, line 118: what happens if you don't have the `Select`s here? I wouldn't have expected those to be necessary.

examples/MomentoRedisLambdaExample/MomentoRedisExampleLambdaHandler/src/MomentoRedisExampleLambdaHandler/MomentoCache.cs
```diff
@@ -0,0 +1,188 @@
...
+            ResponseSize: responseSize,
+            DecompressedResponseSize: decompressedResponseSize,
+            ElapsedMillis: responseElapsedMillis,
+            ElapsedMillisWithDecompression: withDecompressionElapsedMillis
+        );
+        Console.WriteLine($"\tReturning get response for {key}: {getResult}");
+        return getResult;
+    }
+    
+    public async void LogMomentoSetResult(List<Task<SetResult>> momentoSetTasks)
```
cprice404, line 155: Since you are only calling this at the very end of your program, and you know that the tasks will all be completed by then, it would be more clear if you realized all the tasks first and then made this function synchronous, with an argument of type `List<SetResult>`.

I think you should be able to do this via `Task.WhenAll`.  We could probably replace your `WaitAll` statements with this:

`IEnumerable<SetResult> setResults = await Task.WhenAll(momentoSetTasks);`

I'm basing that on this post:

https://briancaos.wordpress.com/2021/02/15/c-get-results-from-task-whenall/

General rule of thumb is it is better to have as few async functions as possible. You should only declare a function async if you must do an `await` in it, and usually only if it needs to return a `Task`. If you can find ways to consolidate your `awaits` in your main function(s) so that the helper functions can be synchronous, it will make the code easier to understand.

examples/MomentoRedisLambdaExample/MomentoRedisExampleLambdaHandler/src/MomentoRedisExampleLambdaHandler/MomentoCache.cs
```diff
@@ -0,0 +1,188 @@
...
+            var setResult = await setTask;
+            if (setResult.Response is CacheSetResponse.Error err) {
+                Console.WriteLine("ERROR: " + err.Message);
+            } else if (setResult.Response is CacheSetResponse.Success) {
+                Console.WriteLine("Set in " + setResult.ElapsedMillis);
+            }
+        }
+    }
+    
+    public async void LogMomentoGetResult(List<Task<GetResult>> momentoGetTasks, bool compressValue)
```
cprice404, line 169: same comment here, better to realize this list of tasks first if possible and make this function synchronous.

rishtigupta:
Still need to extract RedisClusterEndpoint and CacheName as parameters. Will do this in the next PR revision

examples/MomentoRedisLambdaExample/MomentoRedisExampleLambdaHandler/src/MomentoRedisExampleLambdaHandler/RedisCache.cs
```diff
@@ -0,0 +1,143 @@
...
+        Console.WriteLine(pong);
+    }
+
+    public async Task<List<Task<RedisSetResult>>> DoRedisSets(int numRedisSetRequests)
+    {
+        Console.WriteLine("Executing Redis Sets");
+        try
+        {
+            List<Task<RedisSetResult>> redisSetTasks = new List<Task<RedisSetResult>>();
+            string sourceData = await Utils.GetSourceData();
```
cprice404, line 26: same comment here about making this function sync and passing this sourcedata in as an arg

examples/MomentoRedisLambdaExample/MomentoRedisExampleLambdaHandler/src/MomentoRedisExampleLambdaHandler/RedisCache.cs
```diff
@@ -0,0 +1,143 @@
...
+        stopWatch.Stop();
+        var duration = stopWatch.ElapsedMilliseconds;
+        // Console.WriteLine($"\tReturning redis set response for {key}: {setResponse}");        
+        Console.WriteLine($"\tReturning redis set response for {key}");        
+        return new RedisSetResult(Response: setResponse, ElapsedMillis: duration);
+    }
+    
+#pragma warning disable CS1998
+    public async Task<List<Task<RedisGetResult>>> DoRedisGets(int numRedisGetRequests, bool compressValue)
+#pragma warning restore CS1998
```
cprice404, line 64: same comment here about making this function sync

examples/MomentoRedisLambdaExample/MomentoRedisExampleLambdaHandler/src/MomentoRedisExampleLambdaHandler/RedisCache.cs
```diff
@@ -0,0 +1,143 @@
...
+            Response: getResponse,
+            ElapsedMillis: responseElapsedMillis,
+            ElapsedMillisWithDecompression: withDecompressionElapsedMillis
+        );
+        // Console.WriteLine($"\tReturning get response for {key}: {getResult}");
+        Console.WriteLine($"\tReturning get response for {key}");
+        return getResult;
+    }
+
+    public async void LogRedisSetResult(List<Task<RedisSetResult>> redisSetTasks)
```
cprice404, line 122: same comment on this function and the next one about realizing the list of tasks prior to calling this and making it sync.

examples/MomentoRedisLambdaExample/MomentoRedisExampleLambdaHandler/src/MomentoRedisExampleLambdaHandler/Utils.cs
```diff
@@ -0,0 +1,42 @@
...
+
+namespace MomentoRedisExampleLambdaHandler;
+
+public class Utils
+{
+    private static IAmazonS3? _s3Client;
+    private static readonly string BucketName = "my-super-secret-bucket";
+    private static readonly string ObjectKey = "super-secret-source-data.json";
+    
+    public static async Task<string> GetSourceData()
```
cprice404, line 12: nice

examples/MomentoRedisLambdaExample/MomentoRedisExampleLambdaHandler/src/MomentoRedisExampleLambdaHandler/Utils.cs
```diff
@@ -0,0 +1,42 @@
+using Amazon.S3;
+using Amazon.S3.Model;
+
+namespace MomentoRedisExampleLambdaHandler;
+
+public class Utils
+{
+    private static IAmazonS3? _s3Client;
+    private static readonly string BucketName = "my-super-secret-bucket";
```
cprice404, line 9: i'm impressed that you got this bucket name :)

cprice404 (approved):
left some comments just related to how to structure async code, just for learning opportunity. not critical.

examples/MomentoRedisLambdaExample/MomentoRedisExampleLambdaHandler/src/MomentoRedisExampleLambdaHandler/Function.cs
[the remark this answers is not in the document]
pgautier404: That `3000` should be in the TimeSpan below the authProvider then in place of the 30 second value you have there. The client timeout is how long the client will wait for an operation to expire before giving up and returning an error. The default TTL for the cache client is the last parameter in the constructor.

PR #458, opened 2023-07-19
feat: add ItemGetTtl operation

pgautier404:


src/Momento.Sdk/Responses/CacheItemGetTtlResponse.cs
```diff
@@ -0,0 +1,102 @@
...
+            value = response.Found.RemainingTtlMillis;
+        }
+
+        /// <summary>
+        /// The value of the ttl.
+        /// </summary>
+        public ulong Value
+        {
+            get => value;
+        }
```
malandis, line 59: Can we expose this as a `TimeSpan` object instead? That is the .NET native type which has all the ergonomic accessors.

tests/Integration/Momento.Sdk.Tests/TtlTest.cs
```diff
@@ -110,4 +110,50 @@ public async Task UpdateTtlAsync_KeyIsStringAndExists_IsSet()
...
+    {
+        var ttlResponse = await client.ItemGetTtlAsync(Utils.NewGuidString(), Utils.NewGuidString());
+        Assert.True(ttlResponse is CacheItemGetTtlResponse.Error, $"Unexpected response: {ttlResponse}");
+        Assert.Equal(MomentoErrorCode.NOT_FOUND_ERROR, ((CacheItemGetTtlResponse.Error)ttlResponse).ErrorCode);
+    }
+
+    [Fact]
+    public async Task ItemGetTtl_EmptyCacheNameError()
+    {
+        var ttlResponse = await client.ItemGetTtlAsync(null, Utils.NewGuidString());
```
malandis, line 155:
```suggestion
        var ttlResponse = await client.ItemGetTtlAsync(null!, Utils.NewGuidString());
```

malandis (commented):
Looks great just two small changes 🙇‍♂️

src/Momento.Sdk/Responses/CacheItemGetTtlResponse.cs
pgautier404: Yes, good catch.

malandis (approved):
🎉

cprice404 (approved):

PR #462, opened 2023-08-02
chore: updated readme and added more stats to loadgen

pratik151192:
Updated readme, minor cleanup, and added more stats to loadgen

### DIFF
--- a/examples/MomentoLoadGen/README.md
+++ b/examples/MomentoLoadGen/README.md
@@ -40,3 +40,11 @@ Within the `MomentoLoadGen` directory you can run:
 # Run example load generator
 MOMENTO_AUTH_TOKEN=<YOUR AUTH TOKEN> dotnet run
 ```
+
+If you make modifications to the code, remember to do a clean otherwise
+the program might not run.
+
+```bash
+dotnet clean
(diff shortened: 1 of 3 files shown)


cprice404 (approved):

examples/MomentoLoadGen/README.md
```diff
@@ -40,3 +40,11 @@ Within the `MomentoLoadGen` directory you can run:
 # Run example load generator
 MOMENTO_AUTH_TOKEN=<YOUR AUTH TOKEN> dotnet run
 ```
+
+If you make modifications to the code, remember to do a clean otherwise
+the program might not run.
+
+```bash
+dotnet clean
```
cprice404, line 48: FWIW I've never had to do this. but it's probably harmless?

pratik151192: yep harmless

cprice404: one thing that always throws me off when working on .NET is that Visual Studio doesn't auto-save the file so if you re-run without explicitly saving then your changes won't get picked up. Any chance that that is what happened to you?

PR #465, opened 2023-08-10
chore: Relax integration test timing assertions

kvcache:
Momento doesn't offer strict read-after-write consistency, despite
often appearing to. That's an implicit test dependency on the
subjective latency of the test runner, however. By relaxing the
expectation to some small window for a get() response to settle,
the tests should be a little more reliable.

I wrapped CacheClient in a new TestCacheClient implementation of
ICacheClient. Code for delegation to the TestCacheClient's client
field was generated by Visual Studio. This new implementation is
the new CacheClientFixture client implementation.

Also minorly updated some docs.

### DIFF
--- a/tests/Integration/Momento.Sdk.Tests/TestCacheClient.cs
+++ b/tests/Integration/Momento.Sdk.Tests/TestCacheClient.cs
@@ -1,0 +1,374 @@
+using System.Collections.Generic;
+using System.Threading.Tasks;
+using Momento.Sdk.Auth;
+using Momento.Sdk.Config;
+using Momento.Sdk.Requests;
+
+namespace Momento.Sdk.Tests;
+
+/// <summary>
+/// Client call wrapper to make test expectations easier to write.
+/// </summary>
+public class TestCacheClient : ICacheClient
+{
+    private readonly CacheClient client;
+
+    public TestCacheClient(IConfiguration config, ICredentialProvider authProvider, TimeSpan defaultTtl)
+    {
+        client = new CacheClient(config, authProvider, defaultTtl);
+    }
+
+    /// <summary>
+    /// Wait for the service to settle any slow replication or intermediate response cache.
+    /// Momento doesn't have strict read-after-write consistency guarantees, but a little
+    /// time should settle it in normal situations.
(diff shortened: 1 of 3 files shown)


tests/Integration/Momento.Sdk.Tests/TestCacheClient.cs
```diff
@@ -0,0 +1,374 @@
...
+        {
+            await Task.Delay(TimeSpan.FromMilliseconds(10));
+            return await action();
+        }))();
+    }
+
+    public Task<CreateCacheResponse> CreateCacheAsync(string cacheName)
+    {
+        return ((ICacheClient)client).CreateCacheAsync(cacheName);
+    }
```
kvcache, line 38: As mentioned in the pr description, these methods are all generated by visual studio.

tests/Integration/Momento.Sdk.Tests/TestCacheClient.cs
```diff
@@ -0,0 +1,374 @@
...
+
+    public Task<FlushCacheResponse> FlushCacheAsync(string cacheName)
+    {
+        return ((ICacheClient)client).FlushCacheAsync(cacheName);
+    }
+
+    public Task<CacheGetResponse> GetAsync(string cacheName, byte[] key)
+    {
+        return QuiesceAndThenRun(() => ((ICacheClient)client).GetAsync(cacheName, key));
+    }
```
kvcache, line 153: This is how you quiesce then run some client code. (this is the only one I tweaked)

pratik151192 (approved):
LGTM!

tests/Integration/Momento.Sdk.Tests/TestCacheClient.cs
cprice404: why is that cast necessary tho

kvcache: > these methods are all generated by visual studio.

probably not necessary, but this is what visual studio will spew out for the next method you tell it to override. So I just figured this would be the lowest friction thing to do - let the code generator generate code.

PR #466, opened 2023-08-14
chore: also quiesce string keys

kvcache:
missed this in the previous change, https://github.com/momentohq/client-sdk-dotnet/pull/465

Also I made the Quiesce function simpler. I forgot that in c# the
calling convention for an async function is not special. The special
part comes into effect for types which are [awaitable](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/language-specification/expressions#12982-awaitable-expressions).
So, in broad and sweeping generalizations, the async label on a
method doesn't do a magic to the calling convention, only to the
body of the method.


mjindalmotiv8 (approved):

cprice404 (approved):
🙈

PR #468, opened 2023-09-01
feat: add topics

nand4011:
Add the initial topics implementation. The entry point is TopicClient, which contains the publish and subscribe methods. Subscribe returns an IAsyncEnumerable<TopicMessage> that can be iterated over to read from the topic.

### DIFF
--- a/src/Momento.Sdk/Internal/ScsTopicClient.cs
+++ b/src/Momento.Sdk/Internal/ScsTopicClient.cs
@@ -21,7 +21,7 @@ public class ScsTopicClientBase : IDisposable
 
     protected readonly CacheExceptionMapper _exceptionMapper;
 
-    public ScsTopicClientBase(IConfiguration config, string authToken, string endpoint)
+    public ScsTopicClientBase(ITopicConfiguration config, string authToken, string endpoint)
     {
         this.grpcManager = new TopicGrpcManager(config, authToken, endpoint);
         this.dataClientOperationTimeout = config.TransportStrategy.GrpcConfig.Deadline;
@@ -49,7 +49,7 @@ internal sealed class ScsTopicClient : ScsTopicClientBase
 {
     private readonly ILogger _logger;
 
-    public ScsTopicClient(IConfiguration config, string authToken, string endpoint)
+    public ScsTopicClient(ITopicConfiguration config, string authToken, string endpoint)
         : base(config, authToken, endpoint)
     {
         this._logger = config.LoggerFactory.CreateLogger<ScsTopicClient>();
--- a/src/Momento.Sdk/Internal/ScsTopicClient.cs
+++ b/src/Momento.Sdk/Internal/ScsTopicClient.cs
@@ -131,7 +131,7 @@ internal sealed class ScsTopicClient : ScsTopicClientBase
         }
 
         var response = new TopicSubscribeResponse.Subscription(
-            token => MoveNextAsync(subscription, token, cacheName, topicName),
+            cancellationToken => MoveNextAsync(subscription, cancellationToken, cacheName, topicName),
             subscription.Dispose);
         return _logger.LogTraceTopicRequestSuccess(RequestTypeTopicPublish, cacheName, topicName,
             response);
--- a/src/Momento.Sdk/Internal/ScsTopicClient.cs
+++ b/src/Momento.Sdk/Internal/ScsTopicClient.cs
@@ -147,7 +147,7 @@ internal sealed class ScsTopicClient : ScsTopicClientBase
 
         try
         {
-            while (await subscription.ResponseStream.MoveNext(cancellationToken).ConfigureAwait(false))
+            while (await subscription.ResponseStream.MoveNext(cancellationToken))
             {
                 var message = subscription.ResponseStream.Current;
 
--- a/src/Momento.Sdk/Internal/ScsTopicClient.cs
+++ b/src/Momento.Sdk/Internal/ScsTopicClient.cs
@@ -1,3 +1,5 @@
+#if NETSTANDARD2_0_OR_GREATER
+
 using System;
 using System.Threading;
 using System.Threading.Tasks;
@@ -197,3 +199,4 @@ internal sealed class ScsTopicClient : ScsTopicClientBase
         return null;
     }
 }
+#endif
--- a/src/Momento.Sdk/Internal/ScsTopicClient.cs
+++ b/src/Momento.Sdk/Internal/ScsTopicClient.cs
@@ -152,8 +152,20 @@ internal sealed class ScsTopicClient : ScsTopicClientBase
                 switch (message.KindCase)
                 {
                     case _SubscriptionItem.KindOneofCase.Item:
-                        _logger.LogTraceTopicMessageReceived("item", cacheName, topicName);
-                        return new TopicMessage.Item(message.Item);
+                        switch (message.Item.Value.KindCase)
+                        {
+                            case _TopicValue.KindOneofCase.Text:
+                                _logger.LogTraceTopicMessageReceived("text", cacheName, topicName);
+                                return new TopicMessage.Text(message.Item);
+                            case _TopicValue.KindOneofCase.Binary:
+                                _logger.LogTraceTopicMessageReceived("binary", cacheName, topicName);
+                                return new TopicMessage.Binary(message.Item);
+                            case _TopicValue.KindOneofCase.None:
+                            default:
+                                _logger.LogTraceTopicMessageReceived("unknown", cacheName, topicName);
+                                break;
+                        }
+                        break;
                     case _SubscriptionItem.KindOneofCase.Discontinuity:
                         _logger.LogTraceTopicMessageReceived("discontinuity", cacheName, topicName);
                         break;
--- a/src/Momento.Sdk/Internal/ScsTopicClient.cs
+++ b/src/Momento.Sdk/Internal/ScsTopicClient.cs
@@ -132,7 +132,7 @@ internal sealed class ScsTopicClient : ScsTopicClientBase
 
         var response = new TopicSubscribeResponse.Subscription(
             token => MoveNextAsync(subscription, token, cacheName, topicName),
-            () => subscription.Dispose());
+            subscription.Dispose);
         return _logger.LogTraceTopicRequestSuccess(RequestTypeTopicPublish, cacheName, topicName,
             response);
     }
--- a/src/Momento.Sdk/Internal/ScsTopicClient.cs
+++ b/src/Momento.Sdk/Internal/ScsTopicClient.cs
@@ -80,6 +80,7 @@ internal sealed class ScsTopicClient : ScsTopicClientBase
     }
 
     private const string RequestTypeTopicPublish = "TOPIC_PUBLISH";
+    private const string RequestTypeTopicSubscribe = "TOPIC_SUBSCRIBE";
 
     private async Task<TopicPublishResponse> SendPublish(string cacheName, string topicName, _TopicValue value)
     {
@@ -121,23 +122,23 @@ internal sealed class ScsTopicClient : ScsTopicClientBase
         AsyncServerStreamingCall<_SubscriptionItem> subscription;
         try
         {
-            _logger.LogTraceExecutingTopicRequest(RequestTypeTopicPublish, cacheName, topicName);
+            _logger.LogTraceExecutingTopicRequest(RequestTypeTopicSubscribe, cacheName, topicName);
             subscription = grpcManager.Client.subscribe(request, new CallOptions());
         }
         catch (Exception e)
         {
-            return _logger.LogTraceTopicRequestError(RequestTypeTopicPublish, cacheName, topicName,
+            return _logger.LogTraceTopicRequestError(RequestTypeTopicSubscribe, cacheName, topicName,
                 new TopicSubscribeResponse.Error(_exceptionMapper.Convert(e)));
         }
 
         var response = new TopicSubscribeResponse.Subscription(
-            cancellationToken => MoveNextAsync(subscription, cancellationToken, cacheName, topicName),
+            cancellationToken => GetNextRelevantMessageFromGrpcStreamAsync(subscription, cancellationToken, cacheName, topicName),
             subscription.Dispose);
-        return _logger.LogTraceTopicRequestSuccess(RequestTypeTopicPublish, cacheName, topicName,
+        return _logger.LogTraceTopicRequestSuccess(RequestTypeTopicSubscribe, cacheName, topicName,
             response);
     }
 
-    private async ValueTask<TopicMessage?> MoveNextAsync(AsyncServerStreamingCall<_SubscriptionItem> subscription,
+    private async ValueTask<TopicMessage?> GetNextRelevantMessageFromGrpcStreamAsync(AsyncServerStreamingCall<_SubscriptionItem> subscription,
         CancellationToken cancellationToken, string cacheName, string topicName)
     {
         if (cancellationToken.IsCancellationRequested)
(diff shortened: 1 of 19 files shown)


src/Momento.Sdk/Internal/ScsTopicClient.cs
```diff
@@ -0,0 +1,183 @@
...
+            subscription = grpcManager.Client.subscribe(request, new CallOptions());
+        }
+        catch (Exception e)
+        {
+            return _logger.LogTraceTopicRequestError(RequestTypeTopicPublish, cacheName, topicName,
+                new TopicSubscribeResponse.Error(_exceptionMapper.Convert(e)));
+        }
+
+        var response = new TopicSubscribeResponse.Subscription(
+            token => MoveNextAsync(subscription, token, cacheName, topicName),
```
nand4011, line 132: I'm passing in closures here so I can encapsulate the grpc message handling code.

tests/Integration/Momento.Sdk.Tests/TopicTest.cs
```diff
@@ -0,0 +1,244 @@
...
+                    break;
+                }
+            }
+
+            subscription.Dispose();
+            return messageCount;
+        });
+        
+        // Send a few values to skip over to ensure the subscription received the messages under test
+        for (var i = 0; i < 5; i++)
```
nand4011, line 130: This feels hacky but I was having a hard time getting the subscription to consistently receive the first message.

cprice404: Have you tried just putting a sleep here before you start publishing?

tests/Integration/Momento.Sdk.Tests/TopicTest.cs
```diff
@@ -0,0 +1,236 @@
...
+            "two",
+            "three",
+            "four",
+            "five"
+        };
+        
+        var subscribeResponse = await topicClient.SubscribeAsync(cacheName, topicName);
+        Assert.True(subscribeResponse is TopicSubscribeResponse.Subscription,
+            $"Unexpected response: {subscribeResponse}");
+        var subscription = (TopicSubscribeResponse.Subscription)subscribeResponse;
```
cprice404, line 167: Is this what we are expecting the end user code to look like? They will have to cast it?

I know that with the cache methods we were doing like `instanceof Subscription newVarName` and then it would smart cast. If there's not going to be a similar way for users to consume this then we may want to tweak how we are exposing the API, I don't want users to have to cast in the happy path.

tests/Integration/Momento.Sdk.Tests/TopicTest.cs
```diff
@@ -0,0 +1,236 @@
...
+        var testTask = Task.Run(async () =>
+        {
+            var messageCount = 0;
+            await foreach (var message in subscription)
+            {
+                Assert.NotNull(message);
+                Assert.True(message is TopicMessage.Item, $"Unexpected message: {message}");
+                if (((TopicMessage.Item)message).ValueString!.Equals(skipMessage)) continue;
+                
+                Assert.Equal(valuesToSend[messageCount], ((TopicMessage.Item)message).ValueString);
```
cprice404, line 178: same here.

i don't mind casting in tests but I don't want it to be the UX.

If this pattern is just for tests then it might be worth adding an example `main` that shows what you are expecting it to look for users that are consuming it.

The number 1 thing that we have to get correct before we merge this PR is the user-facing API. Everything else can be tweaked/improved/cleaned up later - so you should make sure there is something in the PR that clearly illustrates what you are expecting that to look like for a consumer.

kvcache:
is this missing an assembly reference?
```
+++ b/tests/Integration/Momento.Sdk.Tests/Momento.Sdk.Tests.Integration.csproj
+  <ItemGroup>
+    <Reference Include="Microsoft.Bcl.AsyncInterfaces, Version=7.0.0.0, Culture=neutral, PublicKeyToken=cc7b13ffcd2ddd51" />
+  </ItemGroup>
```

tests/Integration/Momento.Sdk.Tests/TopicTest.cs
nand4011: That's a holdover from my original testing. My documentation uses variable renaming but I wasn't thinking about it with the tests. I'll change it.

src/Momento.Sdk/Responses/TopicMessage.cs
```diff
@@ -0,0 +1,93 @@
...
+
+        /// <summary>
+        /// Constructs an Item from an internal _TopicItem
+        /// </summary>
+        /// <param name="topicItem">Containing the binary or string value.</param>
+        public Item(_TopicItem topicItem)
+        {
+            _value = topicItem.Value;
+            TopicSequenceNumber = topicItem.TopicSequenceNumber;
+        }
```
kvcache, line 48: so, oneofs in protocol buffers are a little different for each language, but they always expect some kind of matching before accessing a property. As-is, a user can't tell the difference between a published empty string and a binary message.
We should provide a way for users to be able to tell the difference, possibly similar to the way protocol buffers does it (that's where we got the go matching pattern), between an Error, a Discontinuity, a Text value and a Binary value. Here is the documentation for inspiration: https://learn.microsoft.com/en-us/dotnet/architecture/grpc-for-wcf-developers/protobuf-any-oneof#oneof

```
        public Item(_TopicItem topicItem)
        {
            switch (topicItem.Value.KindCase)
            {
                case _TopicValue.KindOneofCase.Text:
                    {
                        // The only populated field in this message's value is Text.
                        // Let's vend it to the user that way instead of leaving them to guess?
                        break;
                    }
                case _TopicValue.KindOneofCase.Binary:
                    {
                        // The only populated field in this message's value is Binary.
                        break;
                    }   
                case _TopicValue.KindOneofCase.None:
                    {
                        // Skip message? It's not useful, and the server did something bad
                        break;
                    }
                default:
                    {
                        // Skip message? It's not useful, as it holds something undefined
                        break;
                    }
            }
            _value = topicItem.Value;
            TopicSequenceNumber = topicItem.TopicSequenceNumber;
        }
```

src/Momento.Sdk/Responses/TopicMessage.cs
```diff
@@ -0,0 +1,93 @@
...
+        
+        /// <summary>
+        /// The number of this message in the topic sequence.
+        /// </summary>
+        public ulong TopicSequenceNumber { get; }
+
+        /// <summary>
+        /// The binary value of this message, if present.
+        /// </summary>
+        public byte[] ValueByteArray => _value.Binary.ToByteArray();
```
kvcache, line 58: It seems unnecessarily costly to charge a user a memcpy of the message every time this property is accessed, no?

tests/Integration/Momento.Sdk.Tests/TopicTest.cs
```diff
@@ -0,0 +1,236 @@
...
+            new byte[] { 0x02 },
+            new byte[] { 0x03 },
+            new byte[] { 0x04 },
+            new byte[] { 0x05 }
+        };
+
+        var subscribeResponse = await topicClient.SubscribeAsync(cacheName, topicName);
+        Assert.True(subscribeResponse is TopicSubscribeResponse.Subscription,
+            $"Unexpected response: {subscribeResponse}");
+        var subscription = (TopicSubscribeResponse.Subscription)subscribeResponse;
```
kvcache, line 103: Yeah, casting being the pattern is not great. I think you should consider:
* Make TopicSubscribeResponse the return type.
* Make TopicSubscribeResponse a simple class.
* Un-nest response types.
* Compose 2 things in the TopicSubscribeResponse:
  1. An enum `TopicSubscribeResponse.Kind`
  2. A field for each response kind (like protocol buffers, sound familiar?)
* Add a couple properties to TopicSubscribeResponse:
  1. Kind - documented so people can choose to switch on it if they want to avoid the throws below...
  2. Subscription - gives you the Subscription or throws a good error about the error.
  3. Error - gives you the Error or throws a good error about how this was not an error.

src/Momento.Sdk/Internal/ScsTopicClient.cs
```diff
@@ -0,0 +1,183 @@
...
+                var message = subscription.ResponseStream.Current;
+
+                switch (message.KindCase)
+                {
+                    case _SubscriptionItem.KindOneofCase.Item:
+                        _logger.LogTraceTopicMessageReceived("item", cacheName, topicName);
+                        return new TopicMessage.Item(message.Item);
+                    case _SubscriptionItem.KindOneofCase.Discontinuity:
+                        _logger.LogTraceTopicMessageReceived("discontinuity", cacheName, topicName);
+                        break;
```
kvcache, line 159: this should probably be returning a subscription item so users can be aware of missing messages.

kvcache (commented):
Took a short look and poked at it in visual studio. Nice docs, and I like how we got to `await foreach (var message in subscription)`

src/Momento.Sdk/Responses/TopicMessage.cs
nand4011: I agree with that. I was trying to follow the pattern set by the cache methods, where the value is derived from the same ByteString regardless of whether it is text or binary. Since the difference is more explicit here I'm in favor of making them separate types.

tests/Integration/Momento.Sdk.Tests/TopicTest.cs
nand4011: Casting is not the normal pattern outside of testing. In use it would look more like:
```
if (subscribeResponse is TopicSubscribeResponse.Subscription subscription)
{
      await foreach (var message in subscription)
      {
          switch (message)
          {
              case TopicMessage.Binary binary:
                  break;
              case TopicMessage.Text text:
                  break;
              case TopicMessage.Discontinuity discontinuity:
                  break;
              case TopicMessage.Error:
                  break;
          }
      }
} else if (subscribeResponse is TopicSubscribeResponse.Error error)
{
      // error handling
}
```
That's assuming we split text and binary messages and add discontinuity.

src/Momento.Sdk/Internal/ScsTopicClient.cs
nand4011: I was going to add Discontinuity as a separate item in a second pass, but I can add it now if if won't add too much complexity to the pr.

tests/Integration/Momento.Sdk.Tests/TopicTest.cs
nand4011: @cprice404 Is there a good place to add an example like this without releasing and adding it to the examples directory? I can make something separate if we want realistic looking code before the examples are ready.

kvcache: Oh yeah, that's pretty cool. It looks like a type switch + lexical cast. I couldn't tell that's how it was going :-). Flattening the value cases seems reasonable to me too.

nand4011:
ubuntu-latest and windows-latest seem to hang forever when running the integration tests, while macos-latest passes. The tests run locally on mac os and amazon linux. Continuing to troubleshoot.

nand4011:
This could be a problem:
`C:\Users\runneradmin\.nuget\packages\microsoft.bcl.asyncinterfaces\7.0.0\buildTransitive\net461\Microsoft.Bcl.AsyncInterfaces.targets(4,5): warning : Microsoft.Bcl.AsyncInterfaces 7.0.0 doesn't support net461 and has not been tested with it. Consider upgrading your TargetFramework to net462 or later. You may also set <SuppressTfmSupportBuildWarnings>true</SuppressTfmSupportBuildWarnings> in the project file to ignore this warning and attempt to run in this unsupported configuration at your own risk. [D:\a\client-sdk-dotnet\client-sdk-dotnet\src\Momento.Sdk\Momento.Sdk.csproj]`

src/Momento.Sdk/TopicClient.cs
```diff
@@ -0,0 +1,84 @@
...
+{
+    private readonly ScsTopicClient scsTopicClient;
+    
+    
+    /// <summary>
+    /// Client to perform operations against Momento topics.
+    /// </summary>
+    /// <param name="config">Configuration to use for the transport, retries, middlewares. See <see cref="Configurations"/> for out-of-the-box configuration choices, eg <see cref="Configurations.Laptop.Latest"/></param>
+    /// <param name="authProvider">Momento auth provider.</param>
+    public TopicClient(IConfiguration config, ICredentialProvider authProvider)
```
cprice404, line 26: We do not want to use the CacheClient Configuration object here. We need a new TopicConfiguration interface/object with the minimal surface area.

cprice404 (changes requested):

src/Momento.Sdk/Responses/TopicMessage.cs
cprice404: @kvcache so do you really think that's the right API here? To force users to always have a switch statement in order to read anything at all off of the subscription, even for the simplest use cases?

that feels overbearing and in the other SDKs I think we have been trying to avoid forcing that on them.

tests/Integration/Momento.Sdk.Tests/TopicTest.cs
cprice404: re: how to show examples, generally I will just add a throwaway file with the example code, and remove it before the PR is merged. Or add a temporary dependency on `../blah` in the Examples project and include an example there, and stash it before the final PR merge.

or pasting an example into github like you did is fine too.

src/Momento.Sdk/Internal/ScsTopicClient.cs
cprice404: yes we need to get this out the door, we will add those later

src/Momento.Sdk/Responses/TopicMessage.cs
cprice404: @kvcache random thought:

In the JS SDK we implement subscriptions by allowing the user to pass callbacks for OnItem and OnError. Presumably in the future there might also be OnDiscontinuity.

We did that in JS SDK because that's how the grpc streaming API was exposed.

In golang we did it iterator-style because that's how the grpc streaming API was exposed.

But would it be crazy, in .NET, to do it the callback way?

We could encapsulate the foreach loop inside the SDK in an async function, and allow the user to pass us callbacks for OnBinary, OnText, OnDiscontinuity.

Then they wouldn't have to have a switch statement.

It seems like it might be a better API IMO, with two questions in my mind:

1. Is it going to be obvious to them that the overloaded signatures of `publish` will map to the specific `OnBinary`/`OnText` callbacks? (This issue is relevant now, even if we stick with the iterator-style API.)
2. If they need to do anything that has to do with ordering between OnText and OnDiscontinuity, having those separated into two user-provided lambdas might make it harder for them to manage/share state across the two callbacks? maybe not though, they could close over it or use a class that had the two callback functions?

kvcache: @cprice404 this is inside the Item class - I do think it is reasonable to expect Momento developers to switch to do the needful with each message on the subscription :-).

Separately, I do not think all users should be forced to use pattern matching if they are of the exceptional persuasion. I tried to suggest that we should have accessors available that throw if the message is not of the required disposition. I think the item type should BOTH [1] have a switch pattern available AND [2] have exceptional typed accessors available.

kvcache: oh dude, you sparked a memory: I think we might want to think about using `delegate`s for this. You hit on something interesting here.

kvcache:
```c#
class SubscriptionActions
{
    public SubscriptionActions(TextHandler? onText = null, BinaryHandler? onBinary = null)
    {
        OnTextMessage += Log;
        OnBinaryMessage += Log;

        if (onText != null)
        {
            OnTextMessage += onText;
        }
        if (onBinary != null)
        {
            OnBinaryMessage += onBinary;
        }
    }

    public event TextHandler OnTextMessage;
    public event BinaryHandler OnBinaryMessage;

    private void Log<T>(T message)
    {
    }
}

class Lol
{
    SubscriptionActions MakeSubscriptionActionsThatJustPrintTextMessagesToTheConsole()
    {
        return new SubscriptionActions(onText: Console.WriteLine);
    }
}
```

kvcache: Hmm, as nice as this _could_ be, I think `await foreach` is too ergonomic to ignore.

tests/Integration/Momento.Sdk.Tests/TopicTest.cs
nand4011: The rewritten tests have a consume function that looks much more like what standard use is: https://github.com/momentohq/client-sdk-dotnet/blob/9ad31e5454674c670913f3feca68d64eddd7cf11/tests/Integration/Momento.Sdk.Tests/TopicTest.cs#L171

cprice404 (approved):

PR #476, opened 2023-09-07
feat: add auth client

pgautier404:
This commit adds an auth client that is currently solely responsible for generating disposable tokens.

### DIFF
--- a/src/Momento.Sdk/Auth/StringMomentoTokenProvider.cs
+++ b/src/Momento.Sdk/Auth/StringMomentoTokenProvider.cs
@@ -36,6 +36,7 @@ public class StringMomentoTokenProvider : ICredentialProvider
         var tokenData = AuthUtils.TryDecodeAuthToken(AuthToken);
         ControlEndpoint = tokenData.ControlEndpoint;
         CacheEndpoint = tokenData.CacheEndpoint;
+        TokenEndpoint = tokenData.TokenEndpoint;
         AuthToken = tokenData.AuthToken;
     }
 
--- a/src/Momento.Sdk/Auth/StringMomentoTokenProvider.cs
+++ b/src/Momento.Sdk/Auth/StringMomentoTokenProvider.cs
@@ -17,6 +17,8 @@ public class StringMomentoTokenProvider : ICredentialProvider
     public string ControlEndpoint { get; private set; }
     /// <inheritdoc />
     public string CacheEndpoint { get; private set; }
+    /// <inheritdoc />
+    public string TokenEndpoint { get; private set; }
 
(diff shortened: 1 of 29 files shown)


src/Momento.Sdk/Internal/AuthGrpcManager.cs
```diff
@@ -0,0 +1,105 @@
...
+    }
+
+}
+
+public class AuthGrpcManager : IDisposable
+{
+    private readonly GrpcChannel channel;
+    public IAuthClient Client { get; }
+
+#if USE_GRPC_WEB
```
anitarua, line 51: is having a web endpoint necessary for dotnet? 
I thought it was in the JS SDK because of the split between web vs nodejs packages

src/Momento.Sdk/Auth/Utils.cs
```diff
@@ -65,6 +68,9 @@ public static TokenAndEndpoints TryDecodeAuthToken(string authToken)
                 return new TokenAndEndpoints(
                     authToken,
                     claims.ControlEndpoint,
+                    claims.CacheEndpoint,
+                    // TODO: assuming legacy tokens will never have a token endpoint. AFAIK, they can't be used
+                    //  to generate disposable tokens anyway, but verify and see if this needs to be fixed.
```
cprice404, line 73: this is fine

src/Momento.Sdk/Auth/StringMomentoTokenProvider.cs
```diff
@@ -17,6 +17,8 @@ public class StringMomentoTokenProvider : ICredentialProvider
     public string ControlEndpoint { get; private set; }
     /// <inheritdoc />
     public string CacheEndpoint { get; private set; }
+    /// <inheritdoc />
+    public string TokenEndpoint { get; private set; }
```
cprice404, line 21: where does this get set?

src/Momento.Sdk/Internal/AuthGrpcManager.cs
cprice404: we actually do support grpc-web in .NET, because some really old versions of the .NET framework (and possibly also Unity) do not support HTTP 2.0.

cprice404 (commented):
just the one possible blocker about the token endpoint, otherwise i think we can ship this

src/Momento.Sdk/Auth/StringMomentoTokenProvider.cs
pgautier404: Good catch, thanks. It gets set in b75af12 :-)

PR #478, opened 2023-09-08
feat: reconnect topic subscriptions on error

nand4011:
Instead of returning an error message, attempt to reconnect to a topic on error by remaking the subscription.

### DIFF
--- a/src/Momento.Sdk/Internal/ScsTopicClient.cs
+++ b/src/Momento.Sdk/Internal/ScsTopicClient.cs
@@ -231,6 +231,7 @@ internal sealed class ScsTopicClient : ScsTopicClientBase
                     case _SubscriptionItem.KindOneofCase.Discontinuity:
                         _logger.LogTraceTopicDiscontinuityReceived(_cacheName, _topicName,
                             message.Discontinuity.LastTopicSequence, message.Discontinuity.NewTopicSequence);
+                        _lastSequenceNumber = message.Discontinuity.NewTopicSequence;
                         break;
                     case _SubscriptionItem.KindOneofCase.Heartbeat:
                         _logger.LogTraceTopicMessageReceived("heartbeat", _cacheName, _topicName);
--- a/src/Momento.Sdk/Internal/ScsTopicClient.cs
+++ b/src/Momento.Sdk/Internal/ScsTopicClient.cs
@@ -119,11 +119,11 @@ internal sealed class ScsTopicClient : ScsTopicClientBase
             request.ResumeAtTopicSequenceNumber = resumeAtTopicSequenceNumber.Value;
         }
 
-        AsyncServerStreamingCall<_SubscriptionItem> subscription;
+        SubscriptionWrapper subscriptionWrapper;
         try
         {
             _logger.LogTraceExecutingTopicRequest(RequestTypeTopicSubscribe, cacheName, topicName);
-            subscription = grpcManager.Client.subscribe(request, new CallOptions());
+            subscriptionWrapper = new SubscriptionWrapper(grpcManager, cacheName, topicName, _exceptionMapper, _logger);
         }
         catch (Exception e)
         {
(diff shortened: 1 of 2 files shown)


src/Momento.Sdk/Internal/ScsTopicClient.cs
```diff
@@ -132,72 +132,125 @@ private async Task<TopicPublishResponse> SendPublish(string cacheName, string to
...
-            subscription.Dispose);
+            cancellationToken => subscriptionWrapper.GetNextRelevantMessageFromGrpcStreamAsync(cancellationToken),
+            subscriptionWrapper.Dispose);
         return _logger.LogTraceTopicRequestSuccess(RequestTypeTopicSubscribe, cacheName, topicName,
             response);
     }
 
-    private async ValueTask<TopicMessage?> GetNextRelevantMessageFromGrpcStreamAsync(AsyncServerStreamingCall<_SubscriptionItem> subscription,
-        CancellationToken cancellationToken, string cacheName, string topicName)
+    private class SubscriptionWrapper : IDisposable
```
nand4011, line 141: I didn't do this since I wanted to keep the pr surface as small as possible, but what are people's thoughts on making this the IAsyncEnumerator that the subscription response returns?
https://github.com/momentohq/client-sdk-dotnet/blob/main/src/Momento.Sdk/Responses/TopicSubscribeResponse.cs#L67
That would mean the topic is only actually subscribed to while the enumerator is in use, but it feels like a more simple layer to encapsulate this grpc logic.

src/Momento.Sdk/Internal/ScsTopicClient.cs
```diff
@@ -132,72 +132,125 @@ private async Task<TopicPublishResponse> SendPublish(string cacheName, string to
...
+                    if (sdkException.ErrorCode == MomentoErrorCode.CANCELLED_ERROR)
+                    {
+                        break;
+                    }
+                    //TODO: Are there other errors that we should break for?
+
+                    _logger.LogTraceTopicSubscriptionError(_cacheName, _topicName, sdkException);
+
+                    Dispose();
+                    _subscription = Subscribe();
```
nand4011, line 205: The go sdk checks the connection state before reconnecting. I don't do that here because grpc web doesn't have a connection state to check, but we could conditionally try it.

kvcache: Why do you instantly fire the reconnect and then later await 5 seconds before using it? Shouldn't those go the other way around?

src/Momento.Sdk/Internal/ScsTopicClient.cs
```diff
@@ -132,72 +132,125 @@ private async Task<TopicPublishResponse> SendPublish(string cacheName, string to
...
                             default:
-                                _logger.LogTraceTopicMessageReceived("unknown", cacheName, topicName);
+                                _logger.LogTraceTopicMessageReceived("unknown", _cacheName, _topicName);
                                 break;
                         }
 
                         break;
                     case _SubscriptionItem.KindOneofCase.Discontinuity:
-                        _logger.LogTraceTopicMessageReceived("discontinuity", cacheName, topicName);
+                        _logger.LogTraceTopicMessageReceived("discontinuity", _cacheName, _topicName);
```
kvcache, line 233: in case of discontinuity you should update _lastSequenceNumber to the [`new_topic_sequence`](https://github.com/momentohq/client-protos/blob/main/proto/cachepubsub.proto#L136C10-L136C28) number.

src/Momento.Sdk/Internal/ScsTopicClient.cs
eaddingtonwhite: If that feels more natural for DotNet users then I think be cool to start with that. I'm not the most experienced in dotnet though be good get @kvcache and also maybe ask a few of our customers in slack who use dotnet what they think. Can DM you with few good targets. Also ok w/ me starting here and fast following w/ wrapper or change to add this.

src/Momento.Sdk/Internal/ScsTopicClient.cs
eaddingtonwhite: Also when doing trace log would it be good to include in trace the value of `Discontinuity` token we got?

src/Momento.Sdk/Internal/ScsTopicClient.cs
nand4011: 🤦 I must have mixed those up when I was doing some testing. I'll switch that around.

src/Momento.Sdk/Internal/ScsTopicClient.cs
nand4011: There will be no user facing change, but it would make the subscription response object cleaner.

kvcache (approved):
looks reasonable to me

PR #479, opened 2023-09-08
chore: disposable token example

pgautier404:


examples/DisposableTokens/Program.cs
```diff
@@ -0,0 +1,42 @@
...
+    new DisposableToken.CachePermission(
+        CacheRole.ReadOnly,
+        CacheSelector.ByName("topsecret")
+    ),
+    new DisposableToken.TopicPermission(
+        TopicRole.PublishSubscribe,
+        CacheSelector.ByName("cache"),
+        TopicSelector.ByName("example-topic")
+    )
+});
```
cprice404, line 27: in a subsequent PR it would be good to add another example (can just be later in this same file) that uses the shorthand sugar APIs in `DisposableTokenScopes`.

cprice404 (approved):

PR #481, opened 2023-09-11
chore: add preprocessor directoves for unity build

pgautier404:


src/Momento.Sdk/Auth/EnvMomentoTokenProvider.cs
```diff
@@ -17,8 +17,11 @@ public class EnvMomentoTokenProvider : ICredentialProvider
     public string ControlEndpoint { get; private set; }
     /// <inheritdoc />
     public string CacheEndpoint { get; private set; }
+
+    #if !BUILD_FOR_UNITY
```
cprice404, line 21: I think it's okay to leave this in place regardless of the build target? Just want to try to minimize the number of places we have to keep track of this, especially in the middle of a class/file.

src/Momento.Sdk/Internal/Utils.cs
```diff
@@ -3,7 +3,10 @@
 using System.Collections.Generic;
 using System.Linq;
 using System.Text;
+
+#if !BUILD_FOR_UNITY
 using Momento.Sdk.Exceptions;
+#endif
```
cprice404, line 9: Same comment for this file - if it seems harmless to leave this intact for all builds I think it would be nice

cprice404 (commented):
Lgtm thanks for doing this - just a couple questions about whether we can leave the creds and error handling in place so we have fewer conditional build things, but let me know if you think that's a bad idea.

src/Momento.Sdk/Auth/EnvMomentoTokenProvider.cs
pgautier404: Sure, that makes sense. I'll sweep through and remove relevant directives.

nand4011 (approved):

PR #482, opened 2023-09-11
chore: add topic client configurations

pgautier404:


nand4011 (commented):
The other configs don't have V1s yet in case we want to change things. LGTM other than that.

pgautier404:
> The other configs don't have V1s yet in case we want to change things. LGTM other than that.

Ah, ok, thanks. I'll remove them for consistency and so we aren't breaking promises later.

nand4011 (approved):

cprice404 (approved):

PR #484, opened 2023-09-11
chore: add topics and disposable token code snippets for dev docs

anitarua:
Addresses [#453](https://github.com/momentohq/dev-eco-issue-tracker/issues/453) by adding code snippet for `GenerateDisposableToken` that will be pulled into the dev docs. This snippet can be used [here](https://docs.momentohq.com/develop/authentication/tokens) and [here](https://docs.momentohq.com/develop/api-reference/auth#generatedisposabletoken-api).

Addresses [#459](https://github.com/momentohq/dev-eco-issue-tracker/issues/459) by adding code snippets for instantiating a topic client, publishing to a topic, and subscribing to a topic. These examples will be used [on this page](https://docs.momentohq.com/develop/api-reference/topics).

### DIFF
--- a/examples/DocExampleApis/Program.cs
+++ b/examples/DocExampleApis/Program.cs
@@ -140,38 +140,71 @@ public class Program
 
     public static async Task Example_API_GenerateDisposableToken(IAuthClient authClient)
     {
-        var scope = new DisposableTokenScope(Permissions: new List<DisposableTokenPermission>
-        {
-            new DisposableToken.CacheItemPermission(
-                CacheRole.ReadWrite,
-                CacheSelector.ByName("cache"),
-                CacheItemSelector.AllCacheItems
-            ),
-            new DisposableToken.CachePermission(
-                CacheRole.ReadOnly,
-                CacheSelector.ByName("topsecret")
-            ),
-            new DisposableToken.TopicPermission(
-                TopicRole.PublishSubscribe,
-                CacheSelector.ByName("cache"),
-                TopicSelector.ByName("example-topic")
-            )
-        });
-        var tokenResponse = await authClient.GenerateDisposableTokenAsync(
-            scope,
-            ExpiresIn.Minutes(5)
+        // Generate a disposable token with read-write access to a specific key in one cache
+        var oneKeyOneCacheToken = await authClient.GenerateDisposableTokenAsync(
+            DisposableTokenScopes.CacheKeyReadWrite("squirrels", "mo"),
+            ExpiresIn.Minutes(30)
         );
-
-        if (tokenResponse is GenerateDisposableTokenResponse.Success token)
+        if (oneKeyOneCacheToken is GenerateDisposableTokenResponse.Success token1)
         {
-            Console.WriteLine("The generated disposable token is: " + token.AuthToken);
-            Console.WriteLine("The token endpoint is: " + token.Endpoint);
-            Console.WriteLine("The token expires at (epoch timestamp): " + token.ExpiresAt.Epoch());
+            
+            // logging only a substring of the tokens, because logging security credentials is not advisable :)
+            Console.WriteLine("The generated disposable token starts with: " + token1.AuthToken.Substring(0, 10));
+            Console.WriteLine("The token expires at (epoch timestamp): " + token1.ExpiresAt.Epoch());
         }
-        else if (tokenResponse is GenerateDisposableTokenResponse.Error err)
+        else if (oneKeyOneCacheToken is GenerateDisposableTokenResponse.Error err)
         {
             Console.WriteLine("Error generating disposable token: " + err.Message);
         }
+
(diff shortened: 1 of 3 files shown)


examples/DocExampleApis/Program.cs
```diff
@@ -122,4 +137,101 @@ public static async Task Example_API_Delete(CacheClient cacheClient)
...
+            new DisposableToken.TopicPermission(
+                TopicRole.PublishSubscribe,
+                CacheSelector.ByName("cache"),
+                TopicSelector.ByName("example-topic")
+            )
+        });
+        var tokenResponse = await authClient.GenerateDisposableTokenAsync(
+            scope,
+            ExpiresIn.Minutes(5)
+        );
```
pgautier404, line 163: This is 100% the fault of the example script I threw together, but I think this approach of creating a custom scope with multiple permissions is possibly over-complicated for the API examples. The Javascript example uses the `DisposableTokenScopes` convenience methods to simplify this process and I think we should probably follow that example. So, for example, the selected code would omit the construction of the `scope` variable and use a predefined scope instead:

```
var tokenResponse = await client.GenerateDisposableTokenAsync(
    DisposableTokenScopes.CacheKeyReadWrite("squirrels", "mo"),
    ExpiresIn.Minutes(5)
);
```

The Javascript example shows off 4 of the predefined scopes, and I think it would be good to show multiple of the `DisposableTokenScopes` here too.

Sorry for causing the extra work and for the over-verbose example code. I'm going to either add a second disposable token example program using the predefined scopes or just add an example of each approach in the one program soon so users can study and compare both approaches, but for these docs I think the simplest approach is the best to show.

anitarua: No worries! Easy enough to fix and better to be consistent with the existing examples for sure. I'll use predefined scopes to address the other 3 examples in the JS version as well.

examples/DocExampleApis/Program.cs
```diff
@@ -122,4 +137,134 @@ public static async Task Example_API_Delete(CacheClient cacheClient)
...
+            Console.WriteLine("The token expires at (epoch timestamp): " + token2.ExpiresAt.Epoch());
+        }
+        else if (keyPrefixAllCachesToken is GenerateDisposableTokenResponse.Error err)
+        {
+            Console.WriteLine("Error generating disposable token: " + err.Message);
+        }
+
+        // Generate a disposable token with read-only access to all topics in one cache
+        var allTopicsOneCacheToken = await authClient.GenerateDisposableTokenAsync(
+            DisposableTokenScopes.TopicSubscribeOnly(CacheSelector.ByName("squirrel"), TopicSelector.AllTopics),
```
anitarua, line 178: This one can be changed to remove `CacheSelector.ByName` once the additional predefined scope functions are released

pgautier404: Ah, so we're missing an overload that takes `(string, TopicSelector)`. Good catch. Would you mind making a ticket to remind us to audit these functions and add signatures for `(string, XYZSelector)` please?

pgautier404: Thank you for adding those!

pgautier404:
This LGTM! I think @nand4011 should take a look at the Topic examples before you merge, and if he's good with it, :shipit:

examples/DocExampleApis/Program.cs
```diff
@@ -122,4 +137,134 @@ public static async Task Example_API_Delete(CacheClient cacheClient)
...
+       var publishResponse =
+            await topicClient.PublishAsync("test-cache", "test-topic", "test-topic-value");
+        switch (publishResponse)
+        {
+            case TopicPublishResponse.Success:
+                Console.WriteLine("Successfully published message to 'test-topic'");
+                break;
+            case TopicPublishResponse.Error error:
+                throw new Exception($"An error occurred while publishing topic message: {error.ErrorCode}: {error}");
+                break;
```
nand4011, line 228: I don't think we need a break here after the exception. Looks like there are a couple of these.

examples/DocExampleApis/Program.cs
```diff
@@ -122,4 +137,134 @@ public static async Task Example_API_Delete(CacheClient cacheClient)
...
+    public static async Task Example_API_GenerateDisposableToken(IAuthClient authClient)
+    {
+        // Generate a disposable token with read-write access to a specific key in one cache
+        var oneKeyOneCacheToken = await authClient.GenerateDisposableTokenAsync(
+            DisposableTokenScopes.CacheKeyReadWrite("squirrels", "mo"),
+            ExpiresIn.Minutes(30)
+        );
+        if (oneKeyOneCacheToken is GenerateDisposableTokenResponse.Success token1)
+        {
+
```
nand4011, line 150: I don't normally comment about whitespace, but looking good is half the point of examples :-).

nand4011 (approved):

PR #491, opened 2023-09-14
chore: improve example to include predefined scope

pgautier404:


examples/DisposableTokens/Program.cs
```diff
@@ -5,38 +5,85 @@
...
+            throw err.InnerException;
+        }
+        throw new Exception("Unknown response type: " + tokenResponse);
+    }
+
+    private static void PrintTokenInfo(GenerateDisposableTokenResponse.Success token)
+    {
+        Console.WriteLine(
+            "The generated disposable token (truncated): " 
+            + token.AuthToken.Substring(0, 10)
```
nand4011, line 75: It's far too late for .NET, but for the next sdk we make, we should make a token type that overrides toString() so that it only prints in a truncated way.

nand4011 (approved):

anitarua (approved):

examples/DisposableTokens/Program.cs
pgautier404: I'm adding a `ToString()` override to `GenerateDisposableTokenResponse.Success` in a separate PR right now!

PR #493, opened 2023-09-14
chore: add ToString override for disposable token

pgautier404:
This commit overrides `ToString` for the `GenerateDisposableTokenResponse.Success` class. Printing an object of this type now displays a truncated version of the token (e.g., `eyJlbmRwb2...03dVBYWSJ9`).


anitarua (approved):
very handy!

PR #498, opened 2023-09-20
chore: return an error from the subscribe method for a bad stream

nand4011:
Make the internal subscribe method get the first message and check it to ensure the stream can be read. Without this, certain topic errors won't be caught until the user attempts to iterate over the topic stream.

Break out of the subscription when it receives a permission or authentication error. These cannot be recovered from, so there is no point in reconnecting. This covers the case of a token that expires while being used to read a topic.

Enable the auth client topic tests covering cases that should return an error on subscribe.


pgautier404 (approved):
LGTM!

PR #499, opened 2023-09-20
chore: adding tests for composed disposable token scopes

pgautier404:
This commit adds an assortment of tests against `DisposableTokenScope` objects that are composed of multiple permissions to supplement the existing tests which rely on predefined single-permission `DisposableTokenScopes` objects. Similar tests covering topics are being handled in a separate PR that adds token scoping by topic name prefix.

### DIFF
--- a/tests/Integration/Momento.Sdk.Tests/AuthClientCacheTest.cs
+++ b/tests/Integration/Momento.Sdk.Tests/AuthClientCacheTest.cs
@@ -1,11 +1,8 @@
-using System.ComponentModel.DataAnnotations;
-using System.Diagnostics.CodeAnalysis;
-using System.Runtime.InteropServices;
+using System.Collections.Generic;
 using System.Threading.Tasks;
 using Momento.Sdk.Auth;
 using Momento.Sdk.Auth.AccessControl;
 using Momento.Sdk.Config;
-using Newtonsoft.Json.Serialization;
 
 namespace Momento.Sdk.Tests;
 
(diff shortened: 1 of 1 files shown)


tests/Integration/Momento.Sdk.Tests/AuthClientCacheTest.cs
```diff
@@ -553,4 +550,173 @@ public async Task GenerateDisposableCacheKeyPrefixAuthToken_WriteOnly_HappyPath(
...
+        Assert.True(getResponse is CacheGetResponse.Hit);
+        getResponse = await client.GetAsync(cache2Name, "pet-koala");
+        Assert.True(getResponse is CacheGetResponse.Hit);
+        getResponse = await client.GetAsync(cacheName, "dog");
+        AssertPermissionError<CacheGetResponse, CacheGetResponse.Error>(getResponse);
+        getResponse = await client.GetAsync(cache2Name, "dog");
+        AssertPermissionError<CacheGetResponse, CacheGetResponse.Error>(getResponse);
+
+        // delete the second cache
+        Assert.True(await cacheClient.DeleteCacheAsync(cache2Name) is DeleteCacheResponse.Success);
```
nand4011, line 672: LGTM, but these should go in a finally so we don't leak caches if the test fails.

pgautier404: Totally, great point, thanks.

pgautier404: Fixed in 3a6928b.

anitarua (approved):
lgtm!

PR #506, opened 2023-10-26
feat: Add MVI data methods

nand4011:
Add the upsert, delete, and search MVI methods to the vector client.

Add an internal MVI data client and grpc manager.

Update the protos to get the latest MVI changes.

### DIFF
--- a/src/Momento.Sdk/IPreviewVectorIndexClient.cs
+++ b/src/Momento.Sdk/IPreviewVectorIndexClient.cs
@@ -182,6 +182,6 @@ public interface IPreviewVectorIndexClient : IDisposable
     /// }
     /// </code>
     ///</returns>
-    public Task<VectorSearchResponse> SearchAsync(string indexName, IEnumerable<float> queryVector, uint topK = 10,
+    public Task<VectorSearchResponse> SearchAsync(string indexName, IEnumerable<float> queryVector, int topK = 10,
         MetadataFields? metadataFields = null);
 }
--- a/src/Momento.Sdk/IPreviewVectorIndexClient.cs
+++ b/src/Momento.Sdk/IPreviewVectorIndexClient.cs
@@ -110,7 +110,7 @@ public interface IPreviewVectorIndexClient : IDisposable
     /// If an item with the same ID already exists in the index, it will be replaced.
     /// Otherwise, it will be added to the index.
     /// </summary>
-    /// <param name="indexName">The name of the vector index to delete.</param>
+    /// <param name="indexName">The name of the vector index to upsert the items into.</param>
     /// <param name="items">The items to upsert into the index.</param>
     /// <returns>
     /// Task representing the result of the upsert operation. The
@@ -133,9 +133,9 @@ public interface IPreviewVectorIndexClient : IDisposable
         IEnumerable<Item> items);
 
     /// <summary>
-    /// Deletes all items with the given IDs from the index.
+    /// Deletes all items with the given IDs from the index. Returns success if for items that do not exist.
     /// </summary>
-    /// <param name="indexName">The name of the vector index to delete.</param>
+    /// <param name="indexName">The name of the vector index to delete the items from.</param>
     /// <param name="ids">The IDs of the items to delete from the index.</param>
     /// <returns>
     /// Task representing the result of the upsert operation. The
@@ -161,7 +161,7 @@ public interface IPreviewVectorIndexClient : IDisposable
     /// Ranks the vectors according to the similarity metric specified when the
     /// index was created.
     /// </summary>
-    /// <param name="indexName">The name of the vector index to delete.</param>
+    /// <param name="indexName">The name of the vector index to search in.</param>
     /// <param name="queryVector">The vector to search for.</param>
     /// <param name="topK">The number of results to return. Defaults to 10.</param>
     /// <param name="metadataFields">A list of metadata fields to return with each result.</param>
@@ -170,13 +170,13 @@ public interface IPreviewVectorIndexClient : IDisposable
     /// response object is resolved to a type-safe object of one of
     /// the following subtypes:
     /// <list type="bullet">
-    /// <item><description>DeleteItemBatchResponse.Success</description></item>
-    /// <item><description>DeleteItemBatchResponse.Error</description></item>
+    /// <item><description>SearchResponse.Success</description></item>
+    /// <item><description>SearchResponse.Error</description></item>
     /// </list>
     /// Pattern matching can be used to operate on the appropriate subtype.
     /// For example:
     /// <code>
-    /// if (response is DeleteItemBatchResponse.Error errorResponse)
+    /// if (response is SearchResponse.Error errorResponse)
     /// {
     ///     // handle error as appropriate
     /// }
--- a/src/Momento.Sdk/IPreviewVectorIndexClient.cs
+++ b/src/Momento.Sdk/IPreviewVectorIndexClient.cs
@@ -27,14 +27,14 @@ public interface IPreviewVectorIndexClient : IDisposable
     /// is resolved to a type-safe object of one of
     /// the following subtypes:
     /// <list type="bullet">
-    /// <item><description>CreateVectorIndexResponse.Success</description></item>
-    /// <item><description>CreateVectorIndexResponse.AlreadyExists</description></item>
-    /// <item><description>CreateVectorIndexResponse.Error</description></item>
+    /// <item><description>CreateIndexResponse.Success</description></item>
+    /// <item><description>CreateIndexResponse.AlreadyExists</description></item>
+    /// <item><description>CreateIndexResponse.Error</description></item>
     /// </list>
     /// Pattern matching can be used to operate on the appropriate subtype.
     /// For example:
     /// <code>
-    /// if (response is CreateVectorIndexResponse.Error errorResponse)
+    /// if (response is CreateIndexResponse.Error errorResponse)
     /// {
     ///     // handle error as appropriate
     /// }
@@ -58,7 +58,7 @@ public interface IPreviewVectorIndexClient : IDisposable
     /// </list>
     /// </remarks>
     /// </returns>
-    public Task<CreateVectorIndexResponse> CreateIndexAsync(string indexName, long numDimensions, SimilarityMetric similarityMetric = SimilarityMetric.CosineSimilarity);
+    public Task<CreateIndexResponse> CreateIndexAsync(string indexName, long numDimensions, SimilarityMetric similarityMetric = SimilarityMetric.CosineSimilarity);
     
     /// <summary>
     /// Lists all vector indexes.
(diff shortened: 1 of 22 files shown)


src/Momento.Sdk/Messages/Vector/MetadataValue.cs
nand4011, line 1: This is the biggest point I'm not sure about. We need a clean type-safe way of expressing the different types of metadata in the upsert and search methods. Doing this lets them be defined in a clean way, e.g. [here](https://github.com/momentohq/client-sdk-dotnet/pull/506/files#diff-7db89c3bd215b38e31f65e47700f929a9017e33797823027ea188c6032e7c85bR288), but parsing it feels a bit clunky since you need to switch on the different types like I'm doing [here](https://github.com/momentohq/client-sdk-dotnet/pull/506/files#diff-84eb2ba461b9b65ce120838e6c3b39c9f311523bd765307249956d7fcbd552b5R123). I'm open to any suggestions to clean it up.

src/Momento.Sdk/Requests/Vector/MetadataFields.cs
nand4011, line 1: This uses a similar trick to MetadataValue to allow a method to take either a List<string>, or an ALL sentinel value. The downside is it can't take an IEnumerable. I'm also open to suggestions here.

src/Momento.Sdk/PreviewVectorIndexClient.cs
```diff
@@ -7,10 +8,10 @@
 
 namespace Momento.Sdk;
 
-public class PreviewVectorIndexClient: IPreviewVectorIndexClient
+public class PreviewVectorIndexClient : IPreviewVectorIndexClient
 {
     private readonly VectorIndexControlClient controlClient;
-    
+    private readonly VectorIndexDataClient dataClient;
```
nand4011, line 14: There is no data client round robin like in the cache client currently, nor are there eager connections or retries. I assumed we could add those in a future pr if there is demand.

src/Momento.Sdk/Messages/Vector/MetadataValue.cs
cprice404: do you have examples of what user code will look like?

cprice404: at a glance this seems reasonable and what I would have probably started with, but if you have example user code that would help.

src/Momento.Sdk/Requests/Vector/MetadataFields.cs
cprice404: same comment here, looks reasonable to me at first glance but would love to see example code

src/Momento.Sdk/Requests/Vector/MetadataFields.cs
```diff
@@ -0,0 +1,56 @@
...
+    /// </summary>
+    public static readonly AllFields All = new AllFields();
+
+    /// <summary>
+    /// Implicitly convert a list of strings to a MetadataFields. Allows for passing a bare list instead
+    /// of having to explicitly create a MetadataFields object.
+    /// </summary>
+    /// <param name="fields">The fields to look up.</param>
+    /// <returns></returns>
+    public static implicit operator MetadataFields(List<string> fields) => new List(fields);
```
cprice404, line 22: neat, I didn't know you could do that.

You can't put an enumerable as the argument though?

cprice404 (commented):
i didn't review super closely, and definitely want michael to weigh in. but i skimmed and replied to your comments, and it looks reasonable to me from what I saw!

src/Momento.Sdk/Requests/Vector/MetadataFields.cs
nand4011: Nope, it has to be a specific implementation. The best we can do is add convers for all implementations that make sense.

cprice404: i see.

i actually think list is probably fine, especially for MVP. was just curious because I hadn't seen this feature before.

src/Momento.Sdk/Messages/Vector/MetadataValue.cs
nand4011: The linked test code is what the user code would look like for calling search or upsert. Specifically, creating metadata is easy for the different types:

```csharp
var metadata = new Dictionary<string, MetadataValue>
            {
                { "string_key", "string_value" },
                { "long_key", 123 },
                { "double_key", 3.14 },
                { "bool_key", true },
                { "list_key", new List<string> { "a", "b", "c" } },
                { "empty_list_key", new List<string>() }
            };
```

For search hits that have metadata that could be any type you would need to do something like this:
```csharp
foreach (var hit in successResponse.Hits)
            {
                foreach (var metadataPair in hit.Metadata)
                {
                    var field = metadataPair.Key;
                    switch (metadataPair.Value)
                    {
                        case StringValue stringValue:
                            // do stuff
                            break;
                        case LongValue longValue:
                            // do stuff
                            break;
                        case DoubleValue doubleValue:
                            // do stuff
                            break;
                        case BoolValue boolValue:
                            // do stuff
                            break;
                        case StringListValue stringListValue:
                            // do stuff
                            break;
                    }
                }
            }
```

malandis (commented):
Since .NET has namespacing, I'd prefer the request and response names to not be prefixed with `Vector`. Feels redundant. In python we have elided it, but not in typescript since we don't have a nice option for namespacing.

@cprice404 thoughts?

cprice404:
> Since .NET has namespacing, I'd prefer the request and response names to not be prefixed with `Vector`. Feels redundant. In python we have elided it, but not in typescript since we don't have a nice option for namespacing.
> 
> @cprice404 thoughts?

In a greenfield I definitely agree.

The inconsistency bugs me, so if we do it, I think we should be planning to make a consistent version of the existing ones. Which I think can be done, by basically renaming the `CacheFoo` response to `Cache.Foo`, and then making a type alias (possibly deprecated) using the old name so we don't break backward compat.

We had the same convo about Python and I think we decided to do this, but I don't think we've actually come back and updated the Cache/Topic responses yet.

I'm good with making this change for vectors now as long as we at least capture a ticket for updating the Cache/Topic ones and put it on the board.

src/Momento.Sdk/Requests/Vector/VectorIndexItem.cs
```diff
@@ -0,0 +1,50 @@
+using System.Collections.Generic;
+using Momento.Sdk.Messages.Vector;
+
+namespace Momento.Sdk.Requests.Vector;
+
+/// <summary>
+/// A item in a vector index. Contains an ID, the vector, and any associated metadata.
+/// </summary>
+public class VectorIndexItem
```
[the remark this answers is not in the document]
nand4011: I don't think we can since we the metadata dictionary would be compared by reference. I tried making MetadataValue a record for a while but I couldn't overwrite the equals method in ListStringValue to make it compare correctly. Looks like I overlooked adding an equals and hashcode method here. I can add that in.

src/Momento.Sdk/Responses/Vector/SearchHit.cs
```diff
@@ -0,0 +1,96 @@
...
+
+namespace Momento.Sdk.Responses.Vector;
+
+using System.Collections.Generic;
+
+/// <summary>
+/// A hit from a vector search. Contains the ID of the vector, the distance from the query vector,
+/// and any requested metadata.
+/// </summary>
+public class SearchHit
```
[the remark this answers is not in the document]
nand4011: Same here. I have a custom equals/hashcode that properly compares the metadata. I think changing this to a record would prevent SearchHits from being comparable to each other.

src/Momento.Sdk/IPreviewVectorIndexClient.cs
```diff
@@ -90,17 +91,97 @@ public interface IPreviewVectorIndexClient: IDisposable
...
     ///</returns>
-    public Task<DeleteVectorIndexResponse> DeleteIndexesAsync(string indexName);
+    public Task<DeleteIndexResponse> DeleteIndexesAsync(string indexName);
+
+    /// <summary>
+    /// Upserts a batch of items into a vector index.
+    /// If an item with the same ID already exists in the index, it will be replaced.
+    /// Otherwise, it will be added to the index.
+    /// </summary>
+    /// <param name="indexName">The name of the vector index to delete.</param>
```
malandis, line 113:
```suggestion
    /// <param name="indexName">The name of the vector index to upsert the items in.</param>
```

src/Momento.Sdk/IPreviewVectorIndexClient.cs
```diff
@@ -90,17 +91,97 @@ public interface IPreviewVectorIndexClient: IDisposable
...
+    /// {
+    ///     // handle error as appropriate
+    /// }
+    /// </code>
+    ///</returns>
+    public Task<UpsertItemBatchResponse> UpsertItemBatchAsync(string indexName,
+        IEnumerable<Item> items);
+
+    /// <summary>
+    /// Deletes all items with the given IDs from the index.
```
malandis, line 136: Can you remark that if the item ID doesn't exist, then it's not an error?

src/Momento.Sdk/IPreviewVectorIndexClient.cs
```diff
@@ -90,17 +91,97 @@ public interface IPreviewVectorIndexClient: IDisposable
...
+    /// </summary>
+    /// <param name="indexName">The name of the vector index to delete.</param>
```
malandis, line 138:
```suggestion
    /// <param name="indexName">The name of the vector index to delete the items from.</param>
```

src/Momento.Sdk/IPreviewVectorIndexClient.cs
```diff
@@ -90,17 +91,97 @@ public interface IPreviewVectorIndexClient: IDisposable
...
+    /// </code>
+    ///</returns>
+    public Task<DeleteItemBatchResponse> DeleteItemBatchAsync(string indexName, IEnumerable<string> ids);
+
+    /// <summary>
+    /// Searches for the most similar vectors to the query vector in the index.
+    /// Ranks the vectors according to the similarity metric specified when the
+    /// index was created.
+    /// </summary>
+    /// <param name="indexName">The name of the vector index to delete.</param>
```
malandis, line 164:
```suggestion
    /// <param name="indexName">The name of the vector index to search in.</param>
```

src/Momento.Sdk/IPreviewVectorIndexClient.cs
```diff
@@ -90,17 +91,97 @@ public interface IPreviewVectorIndexClient: IDisposable
...
+    /// response object is resolved to a type-safe object of one of
+    /// the following subtypes:
+    /// <list type="bullet">
+    /// <item><description>DeleteItemBatchResponse.Success</description></item>
+    /// <item><description>DeleteItemBatchResponse.Error</description></item>
+    /// </list>
+    /// Pattern matching can be used to operate on the appropriate subtype.
+    /// For example:
+    /// <code>
+    /// if (response is DeleteItemBatchResponse.Error errorResponse)
```
malandis, line 179: Update this section for `Search`

tests/Integration/Momento.Sdk.Tests/VectorIndexDataTest.cs
```diff
@@ -0,0 +1,333 @@
...
+
+    [Fact]
+    public async Task UpsertAndSearch_InnerProduct()
+    {
+        var indexName = $"index-{Utils.NewGuidString()}";
+
+        var createResponse = await vectorIndexClient.CreateIndexAsync(indexName, 2, SimilarityMetric.InnerProduct);
+        Assert.True(createResponse is CreateIndexResponse.Success, $"Unexpected response: {createResponse}");
+
+        try
```
malandis, line 68: This repeated code and cleanup can be refactored to a disposable class, then you can use a using block; see https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/statements/using

malandis (commented):
Looks good; a few minor comments

malandis:
LGTM. I will 👍 once the conflicts are resolved.

malandis (approved):
😎

PR #509, opened 2023-11-01
fix: clean up after MVI control tests

nand4011:
Properly delete indexes in the MVI control tests.

Give intermittently failing TTL test more generous timing and add failure messages.

### DIFF
--- a/tests/Integration/Momento.Sdk.Tests/VectorIndexControlTest.cs
+++ b/tests/Integration/Momento.Sdk.Tests/VectorIndexControlTest.cs
@@ -17,17 +17,22 @@ public class VectorIndexControlTest : IClassFixture<VectorIndexClientFixture>
     {
         var indexName = $"dotnet-integration-{Utils.NewGuidString()}";
         const int numDimensions = 3;
-        
-        var createResponse = await vectorIndexClient.CreateIndexAsync(indexName, numDimensions);
-        Assert.True(createResponse is CreateVectorIndexResponse.Success, $"Unexpected response: {createResponse}");
-        
-        var listResponse = await vectorIndexClient.ListIndexesAsync();
-        Assert.True(listResponse is ListVectorIndexesResponse.Success, $"Unexpected response: {listResponse}");
-        var listOk = (ListVectorIndexesResponse.Success)listResponse;
-        Assert.Contains(listOk.IndexNames, name => name == indexName);
-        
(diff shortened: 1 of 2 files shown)


malandis (approved):

tests/Integration/Momento.Sdk.Tests/VectorIndexControlTest.cs
```diff
@@ -17,30 +17,43 @@ public async Task CreateListDelete_HappyPath()
...
+            var createResponse = await vectorIndexClient.CreateIndexAsync(indexName, numDimensions);
+            Assert.True(createResponse is CreateVectorIndexResponse.Success, $"Unexpected response: {createResponse}");
+
+            var createAgainResponse = await vectorIndexClient.CreateIndexAsync(indexName, numDimensions);
+            Assert.True(createAgainResponse is CreateVectorIndexResponse.AlreadyExists, $"Unexpected response: {createAgainResponse}");
+        }
+        finally
+        {
+            var deleteResponse = await vectorIndexClient.DeleteIndexesAsync(indexName);
+            Assert.True(deleteResponse is DeleteVectorIndexResponse.Success, $"Unexpected response: {deleteResponse}");
```
virratanasangpunth, line 55: I'm not sure if we should assert that delete is successful. This clean-up finally block is best-effort. IMO we should do
```
finally {
  try {
     var deleteResponse = await vectorIndexClient.DeleteIndexesAsync(indexName);
  } catch (e: Exception) {
    // log and continue;
  }

}

```

Otherwise the failure from this finally block can mask the actual failure from the try block above.

cprice404: agreed. also might be good to fast follow this with a higher-order helper function called `withTestVectorIndex` that encapsulates the cleanup so that we have an easy repeatable pattern for it.

nand4011: I can switch to just throwing an exception. We need to avoid a passing test containing a failing delete call.

cprice404 (approved):
i'm fine with merging as-is if that will start the process of fixing things for @virratanasangpunth , and fast follow with the suggested improvements, or we can add a commit.

nand4011:
I'm going to merge and add an encapsulating cleanup function in the data methods pr. I was talking with Michael about it yesterday. We need to make it work without IAsyncDisposable, since that isn't supported in .NET 461.

PR #510, opened 2023-11-02
chore: make sure all assert.true statements include message on failure

anitarua:
Addresses https://github.com/momentohq/oncall-tracker/issues/3051

Updated all `Assert.True` statements to include a message to print upon failure. Also uncomments the flaky test so it's back in now.

### DIFF
--- a/tests/Integration/Momento.Sdk.Tests/AuthClientTopicTest.cs
+++ b/tests/Integration/Momento.Sdk.Tests/AuthClientTopicTest.cs
@@ -423,8 +423,6 @@ public class AuthClientTopicTest : IClassFixture<CacheClientFixture>, IClassFixt
         await GenerateDisposableTopicAuthToken_WriteOnly_CanPublish_Common(writeOnlyTopicClient, messageValue);
     }
 
-    // TODO: this test is flaky and has been breaking build pipelines. Commenting it out until we have time to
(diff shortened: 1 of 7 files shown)


tests/Integration/Momento.Sdk.Tests/AuthClientTopicTest.cs
```diff
@@ -428,15 +428,15 @@ public async Task GenerateDisposableTopicAuthToken_WriteOnly_CanPublish()
 
     // TODO: this test is flaky and has been breaking build pipelines. Commenting it out until we have time to
     //  root cause it. Chris 2023-10-30
```
anitarua, line 430: should I leave this test commented out?

cprice404: nope we should uncomment it please

cprice404: it was only failing due to throttles so the stuff pete is doing should fix that

cprice404 (dismissed):
hero

cprice404:
thanks a bunch for doing this! you can uncomment that one test either in this PR or a follow-up

cprice404 (approved):

PR #517, opened 2023-11-14
feat: Add MVI SearchAndFetchVectors method

nand4011:
Add a new version of the MVI search method that returns the vectors with the hits.

Paramaterize the search integration tests so that they cover both search methods.


malandis (approved):
Very nice!

PR #519, opened 2023-11-15
feat: add mvi list indexes detail

malandis:
Add MVI list indexes detail to IndexInfo object and update integration tests.

### DIFF
--- a/src/Momento.Sdk/Responses/Vector/IndexInfo.cs
+++ b/src/Momento.Sdk/Responses/Vector/IndexInfo.cs
@@ -48,6 +48,7 @@ public class IndexInfo
         // Overflow is standard here since the alternative is modulo arithmetic, which just wraps around the same.
         unchecked
         {
+            // see https://github.com/dotnet/coreclr/blob/master/src/System.Private.CoreLib/shared/System/HashCode.cs#L58
             var hash = 17;
             hash = hash * 23 + Name.GetHashCode();
             hash = hash * 23 + NumDimensions.GetHashCode();
--- a/src/Momento.Sdk/Responses/Vector/IndexInfo.cs
+++ b/src/Momento.Sdk/Responses/Vector/IndexInfo.cs
@@ -1,3 +1,5 @@
+using Momento.Sdk.Requests.Vector;
+
 namespace Momento.Sdk.Responses.Vector;
 
 /// <summary>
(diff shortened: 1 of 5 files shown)


src/Momento.Sdk/Responses/Vector/IndexInfo.cs
```diff
@@ -10,30 +12,52 @@ public class IndexInfo
...
     /// <inheritdoc />
     public override int GetHashCode()
     {
-        return Name.GetHashCode();
+        unchecked
+        {
+            var hash = 17;
+            hash = hash * 23 + Name.GetHashCode();
+            hash = hash * 23 + NumDimensions.GetHashCode();
+            hash = hash * 23 + SimilarityMetric.GetHashCode();
```
eaddingtonwhite, line 53: JC where do 17 and 23 come from? Could we make named constants for them for what they represent? Just not clear to me looking at code why they equal to what they are.

src/Momento.Sdk/Responses/Vector/IndexInfo.cs
```diff
@@ -10,30 +12,52 @@ public class IndexInfo
...
     {
-        return obj is IndexInfo other && Name == other.Name;
+        return obj is IndexInfo other && Name == other.Name && NumDimensions == other.NumDimensions && SimilarityMetric == other.SimilarityMetric;
     }
 
     /// <inheritdoc />
     public override int GetHashCode()
     {
-        return Name.GetHashCode();
+        unchecked
```
eaddingtonwhite, line 48: this seems scary to me seeing this. Maybe just leave comment why we need?

eaddingtonwhite (commented):
@malandis overall looks good. Just 2 nits around adding some context for next dev that comes in here in comments.

src/Momento.Sdk/Responses/Vector/IndexInfo.cs
malandis: These are prime numbers used to combine the hash codes. There is a .NET standard implementation to combine the hash code representations, but we are targeting such old frameworks we don't have access to it, hence have to copy that here. See here https://github.com/dotnet/coreclr/blob/master/src/System.Private.CoreLib/shared/System/HashCode.cs#L58

eaddingtonwhite: I see ok maybe just comment in code linking out to this at least? I would never have figured this out on my own or had context why were doing.

malandis:
> @malandis overall looks good. Just 2 nits around adding some context for next dev that comes in here in comments.

@eaddingtonwhite done

eaddingtonwhite (approved):

PR #526, opened 2024-02-03
feat: update unity zip

nuernber:
Update the MomentoSdkUnity packaging to include only the necessary .dll files, and also add the version number to the final .dll.

Only tested locally on my Windows machine using the powershell and git bash terminals.


cprice404:
CI is failing because the PR needs to be opened from a branch on the momento org. I will pull down this branch and open a duplicate PR.

PR #527, opened 2024-02-06
feat: update unity zip

cprice404:
This is just a duplicate of #526 , filed against a branch in the momentohq repo so that we can let CI run.

- copy needed .dlls for unity
- set dll version
- fix which dlls to include


nand4011 (approved):
Allow listing seems like a better way. 👍

PR #530, opened 2024-02-07
chore: add missing Newtonsoft.Json.dll

cprice404:


nand4011 (approved):

cprice404:
this is just a dupe of #529 , targeted against the momento org instead of a fork.

PR #534, opened 2024-02-23
fix: include exception in log message when eager connection fails; update env vars to MOMENTO_API_KEY

cprice404:
Prior to this commit, if we failed to eagerly connect, we were not logging any information about
the reason. This commit adds the exception to the log. Tested by disabling my network connection
and verified it includes all the relevant info.

Also updates places in example code where we were still using MOMENTO_AUTH_TOKEN for env var names,
now that we have standardized on MOMENTO_API_KEY.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -60,7 +60,7 @@ jobs:
     env:
       # TODO: remove token stored as secret in favor of using a
       # momento-local instance that can be spun up for testing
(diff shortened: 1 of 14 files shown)


kvcache (commented):
The logging update looks great to me, but I do not have any context on the environment variable.

cprice404:
> The logging update looks great to me, but I do not have any context on the environment variable.

yeah i should have done that as 2 PRs. sorry. just couldn't resist fixing it while i was in there.

kvcache (approved):

pratik151192 (approved):

PR #538, opened 2024-02-28
feat: add grpc config options and turn off keepalive for Lambda config"

anitarua:
DRYs up the grpc channel options setting into a single function called by the various grpc managers.
Adds keepalive settings to all prebuilt configs with a `StaticGrpcConfiguration` specified except for Lambda, which should not have keepalives turned on.

### DIFF
--- a/src/Momento.Sdk/Config/TopicConfigurations.cs
+++ b/src/Momento.Sdk/Config/TopicConfigurations.cs
@@ -35,10 +35,12 @@ public class TopicConfigurations
             var finalLoggerFactory = loggerFactory ?? NullLoggerFactory.Instance;
             ITopicTransportStrategy transportStrategy = new StaticTopicTransportStrategy(
                 loggerFactory: finalLoggerFactory,
-                grpcConfig: new StaticGrpcConfiguration(deadline: TimeSpan.FromMilliseconds(15000)),
-                keepAlivePermitWithoutCalls: true,
-                keepAlivePingDelay: TimeSpan.FromMilliseconds(5000),
-                keepAlivePingTimeout: TimeSpan.FromMilliseconds(1000)
+                grpcConfig: new StaticGrpcConfiguration(
+                    deadline: TimeSpan.FromMilliseconds(15000),
+                    keepAlivePermitWithoutCalls: true,
+                    keepAlivePingDelay: TimeSpan.FromMilliseconds(5000),
+                    keepAlivePingTimeout: TimeSpan.FromMilliseconds(1000)
+                )
             );
             return new Laptop(finalLoggerFactory, transportStrategy);
         }
@@ -68,10 +70,12 @@ public class TopicConfigurations
             var finalLoggerFactory = loggerFactory ?? NullLoggerFactory.Instance;
             ITopicTransportStrategy transportStrategy = new StaticTopicTransportStrategy(
                 loggerFactory: finalLoggerFactory,
-                grpcConfig: new StaticGrpcConfiguration(deadline: TimeSpan.FromMilliseconds(15000)),
-                keepAlivePermitWithoutCalls: true,
-                keepAlivePingDelay: TimeSpan.FromMilliseconds(5000),
-                keepAlivePingTimeout: TimeSpan.FromMilliseconds(1000)
+                grpcConfig: new StaticGrpcConfiguration(
+                    deadline: TimeSpan.FromMilliseconds(15000),
+                    keepAlivePermitWithoutCalls: true,
+                    keepAlivePingDelay: TimeSpan.FromMilliseconds(5000),
+                    keepAlivePingTimeout: TimeSpan.FromMilliseconds(1000)
+                )
             );
             return new Mobile(finalLoggerFactory, transportStrategy);
         }
@@ -108,10 +112,12 @@ public class TopicConfigurations
                 var finalLoggerFactory = loggerFactory ?? NullLoggerFactory.Instance;
                 ITopicTransportStrategy transportStrategy = new StaticTopicTransportStrategy(
                     loggerFactory: finalLoggerFactory,
-                    grpcConfig: new StaticGrpcConfiguration(deadline: TimeSpan.FromMilliseconds(1100)),
-                    keepAlivePermitWithoutCalls: true,
-                    keepAlivePingDelay: TimeSpan.FromMilliseconds(5000),
-                    keepAlivePingTimeout: TimeSpan.FromMilliseconds(1000)
+                    grpcConfig: new StaticGrpcConfiguration(
+                        deadline: TimeSpan.FromMilliseconds(1100),
+                        keepAlivePermitWithoutCalls: true,
+                        keepAlivePingDelay: TimeSpan.FromMilliseconds(5000),
+                        keepAlivePingTimeout: TimeSpan.FromMilliseconds(1000)
+                    )
                 );
                 return new Default(finalLoggerFactory, transportStrategy);
             }
@@ -145,10 +151,12 @@ public class TopicConfigurations
                 var finalLoggerFactory = loggerFactory ?? NullLoggerFactory.Instance;
                 ITopicTransportStrategy transportStrategy = new StaticTopicTransportStrategy(
                     loggerFactory: finalLoggerFactory,
-                    grpcConfig: new StaticGrpcConfiguration(deadline: TimeSpan.FromMilliseconds(500)),
-                    keepAlivePermitWithoutCalls: true,
-                    keepAlivePingDelay: TimeSpan.FromMilliseconds(5000),
-                    keepAlivePingTimeout: TimeSpan.FromMilliseconds(1000)
+                    grpcConfig: new StaticGrpcConfiguration(
+                        deadline: TimeSpan.FromMilliseconds(500),
+                        keepAlivePermitWithoutCalls: true,
+                        keepAlivePingDelay: TimeSpan.FromMilliseconds(5000),
+                        keepAlivePingTimeout: TimeSpan.FromMilliseconds(1000)
+                    )
                 );
                 return new LowLatency(finalLoggerFactory, transportStrategy);
             }
--- a/src/Momento.Sdk/Config/TopicConfigurations.cs
+++ b/src/Momento.Sdk/Config/TopicConfigurations.cs
@@ -35,7 +35,10 @@ public class TopicConfigurations
             var finalLoggerFactory = loggerFactory ?? NullLoggerFactory.Instance;
             ITopicTransportStrategy transportStrategy = new StaticTopicTransportStrategy(
                 loggerFactory: finalLoggerFactory,
-                grpcConfig: new StaticGrpcConfiguration(deadline: TimeSpan.FromMilliseconds(15000))
+                grpcConfig: new StaticGrpcConfiguration(deadline: TimeSpan.FromMilliseconds(15000)),
+                keepAlivePermitWithoutCalls: true,
+                keepAlivePingDelay: TimeSpan.FromMilliseconds(5000),
+                keepAlivePingTimeout: TimeSpan.FromMilliseconds(1000)
             );
             return new Laptop(finalLoggerFactory, transportStrategy);
         }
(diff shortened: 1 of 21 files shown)


src/Momento.Sdk/Config/Transport/IGrpcConfiguration.cs
```diff
@@ -42,6 +43,36 @@ public interface IGrpcConfiguration
...
+
+    /// <summary>
+    /// Indicates if it permissible to send keepalive pings from the client without any outstanding streams.
+    /// NOTE: keep-alives are very important for long-lived server environments where there may be periods of time
+    /// when the connection is idle. However, they are very problematic for lambda environments where the lambda
+    /// runtime is continuously frozen and unfrozen, because the lambda may be frozen before the "ACK" is received
+    /// from the server. This can cause the keep-alive to timeout even though the connection is completely healthy.
+    /// Therefore, keep-alives should be disabled in lambda and similar environments.
+    /// </summary>
+    public bool KeepAlivePermitWithoutCalls { get; }
```
anitarua, line 74: Using type `bool` here because I could not get the build to recognize the `System.Net.Http.HttpKeepAlivePingPolicy` enum that's actually used when setting the keepalive option in the `SocketsHttpHandler` in the `Utils.GrpcChannelOptionsFromGrpcConfig` function

src/Momento.Sdk/Config/Transport/StaticTransportStrategy.cs
```diff
@@ -27,13 +34,27 @@ public class StaticGrpcConfiguration : IGrpcConfiguration
...
     {
         Utils.ArgumentStrictlyPositive(deadline, nameof(deadline));
         this.Deadline = deadline;
         this.MinNumGrpcChannels = minNumGrpcChannels;
-        this.GrpcChannelOptions = grpcChannelOptions ?? new GrpcChannelOptions();
+        this.GrpcChannelOptions = grpcChannelOptions;
         this.SocketsHttpHandlerOptions = socketsHttpHandlerOptions ?? new SocketsHttpHandlerOptions();
+        this.KeepAlivePingTimeout = keepAlivePingTimeout ?? System.Threading.Timeout.InfiniteTimeSpan;
+        this.KeepAlivePingDelay = keepAlivePingDelay ?? System.Threading.Timeout.InfiniteTimeSpan;
+        this.KeepAlivePermitWithoutCalls = keepAlivePermitWithoutCalls ?? false;
```
anitarua, line 57: couldn't find a way to optionally set or omit each keepalive setting when creating the `SocketsHttpHandler` in the `Utils.GrpcChannelOptionsFromGrpcConfig` function, so this part will set each one to the config-provided value or to a default value ([delay](https://learn.microsoft.com/en-us/dotnet/api/system.net.http.socketshttphandler.keepalivepingdelay?view=net-6.0#system-net-http-socketshttphandler-keepalivepingdelay) and [timeout](https://learn.microsoft.com/en-us/dotnet/api/system.net.http.socketshttphandler.keepalivepingtimeout?view=net-6.0#system-net-http-socketshttphandler-keepalivepingtimeout) are disabled when set to infinite time span, policy defaults to [WithActiveRequests](https://learn.microsoft.com/en-us/dotnet/api/system.net.http.httpkeepalivepingpolicy?view=net-6.0))

src/Momento.Sdk/Config/Transport/IGrpcConfiguration.cs
```diff
@@ -31,7 +32,7 @@ public interface IGrpcConfiguration
     /// interface for the channel options, which allows modifying specific values but does
     /// not allow the caller to use arbitrary strings to set the channel options.)
     /// </summary>
-    public GrpcChannelOptions GrpcChannelOptions { get; }
+    public GrpcChannelOptions? GrpcChannelOptions { get; }
```
anitarua, line 35: Made this nullable so that the `Utils.GrpcChannelOptionsFromGrpcConfig` could do a null check before setting the message send/receive sizes. If set to a `new GrpcChannelOptions()` object before reaching that function, `MaxReceiveMessageSize` will already be set to 4mb and it becomes ambiguous as to whether the user actually wants to set it to 4mb or if it was just the default.

src/Momento.Sdk/Config/Transport/IGrpcConfiguration.cs
```diff
@@ -42,6 +43,36 @@ public interface IGrpcConfiguration
     /// </remarks>
     public SocketsHttpHandlerOptions SocketsHttpHandlerOptions { get; }
```
anitarua, line 44: the keepalive settings are technically options set on the `SocketsHttpHandler` so they could be put into the [`SocketsHttpHandlerOptions`](https://github.com/momentohq/client-sdk-dotnet/blob/main/src/Momento.Sdk/Config/Transport/SocketsHttpHandlerOptions.cs) object instead if it seems cleaner to do it that way

src/Momento.Sdk/PreviewVectorIndexClient.cs
```diff
@@ -32,7 +32,7 @@ public PreviewVectorIndexClient(IVectorIndexConfiguration config, ICredentialPro
     {
         var loggerFactory = config.LoggerFactory;
         controlClient =
-            new VectorIndexControlClient(loggerFactory, authProvider.AuthToken, authProvider.ControlEndpoint);
+            new VectorIndexControlClient(config, authProvider.AuthToken, authProvider.ControlEndpoint);
```
anitarua, line 35: Piped the vector index config through to the `VectorIndexControlGrpcManager` but I'm not actually entirely sure if we want control clients to enable keepalives. 

They're not enabled in the JS cache control and vector index control clients, but I'd been enabling them in the Go and Python PRs along the way before I realized that.

src/Momento.Sdk/Internal/Utils.cs
```diff
@@ -18,6 +26,42 @@ public static class Utils
     /// </summary>
     public const int DEFAULT_MAX_MESSAGE_SIZE = 5_243_000;
 
+    /// <summary>
+    /// Create a GrpcChannelOptions object from an IGrpcConfiguration object.
+    /// </summary>
+    /// <param name="grpcConfig">The IGrpcConfiguration object specifying underlying grpc options</param>
+    /// <param name="loggerFactory"></param>
+    /// <returns></returns>
+    public static GrpcChannelOptions GrpcChannelOptionsFromGrpcConfig(IGrpcConfiguration grpcConfig, ILoggerFactory loggerFactory) {
```
nand4011, line 35: It feels strange putting this in a generic utils class. Would it make sense to have an abstract GrpcManager for this?

nand4011 (commented):
Thanks for doing this! The repeated grpc code has been bugging me for a while. I have a question about organization, but LGTM otherwise.

src/Momento.Sdk/Internal/Utils.cs
anitarua: For sure, I'll give that a shot

cprice404: lol, @nand4011 , that's a very OO suggestion :) I kind of like this being a pure function since it has nothing to do with any Object's `this`.

this is not a blocker either way for me but I will always choose FP paradigm when I don't need access to member variables for an object.

src/Momento.Sdk/Config/TopicConfigurations.cs
```diff
@@ -135,7 +151,12 @@ public static ITopicConfiguration Latest(ILoggerFactory? loggerFactory = null)
                 var finalLoggerFactory = loggerFactory ?? NullLoggerFactory.Instance;
                 ITopicTransportStrategy transportStrategy = new StaticTopicTransportStrategy(
                     loggerFactory: finalLoggerFactory,
-                    grpcConfig: new StaticGrpcConfiguration(deadline: TimeSpan.FromMilliseconds(500))
+                    grpcConfig: new StaticGrpcConfiguration(
+                        deadline: TimeSpan.FromMilliseconds(500),
+                        keepAlivePermitWithoutCalls: true,
+                        keepAlivePingDelay: TimeSpan.FromMilliseconds(5000),
+                        keepAlivePingTimeout: TimeSpan.FromMilliseconds(1000)
```
cprice404, line 158: as per our previous slack thread it might be nice to centralize the defaults.

src/Momento.Sdk/Config/Transport/IGrpcConfiguration.cs
cprice404: the impedance mismatch you are running into here is that in all of our other SDKs we've modeled the GRPC config as a custom object that we define and own, whereas in this SDK we are just using the .NET type directly. 

In this model, we will want to avoid defining any new fields on our own config classes that overlap with the things available on the .NET type, and just use the built-in .NET ones to configure things. This may mean that you don't need a `GrpcChannelOptionsFromGrpcConfig` function at all, but I'll have to look more closely at what other fields you're using from our bespoke Config object to make modifications to the ChannelOptions object.

src/Momento.Sdk/Config/Transport/IGrpcConfiguration.cs
cprice404: yeah that sounds like the right way to go to me

src/Momento.Sdk/Config/Transport/IGrpcConfiguration.cs
cprice404: bool sounds okay to me, in general I usually try to avoid leaking types from dependency libraries into the public API of my own library. We have already done that with ChannelOptions so it's not a completely black and white thing, but I think a boolean is fine since we are modeling our own config fields here.

src/Momento.Sdk/Config/Transport/StaticTransportStrategy.cs
```diff
@@ -27,13 +34,27 @@ public class StaticGrpcConfiguration : IGrpcConfiguration
...
+        SocketsHttpHandlerOptions? socketsHttpHandlerOptions = null,
+        TimeSpan? keepAlivePingTimeout = null, 
+        TimeSpan? keepAlivePingDelay = null, 
+        bool? keepAlivePermitWithoutCalls = null
+    )
     {
         Utils.ArgumentStrictlyPositive(deadline, nameof(deadline));
         this.Deadline = deadline;
         this.MinNumGrpcChannels = minNumGrpcChannels;
-        this.GrpcChannelOptions = grpcChannelOptions ?? new GrpcChannelOptions();
```
cprice404, line 35: I think maybe what you want to do is make this `?? DefaultGrpcChannelOptions()`, and then define that function instead of having a `FromGrpcConfig` function? It's the same issue I mentioned earlier where this SDK is a little different because we are using the .NET types directly. So we shouldn't need to create our own config types to model anything that can be expressed inside of the .NET GrpcChannelOptions type, because we're already exposing it.

src/Momento.Sdk/Config/Transport/StaticTransportStrategy.cs
cprice404: i think this will shake out more cleanly when you get rid of `FromGrpcConfig`and put these new config options into the SocketHandlerOptions section as you suggested above?

src/Momento.Sdk/PreviewVectorIndexClient.cs
cprice404: yep i think there's no reason control and data clients shouldn't be configured consistently for now

cprice404 (commented):
looking good, thanks! i had a few suggestions, and after you take another pass we should definitely get @malandis to sign off on this before it goes in :)

src/Momento.Sdk/Config/Transport/IGrpcConfiguration.cs
malandis: re: build failures -- .NET has portability issues because different runtimes support different HttpHandlers. For this reason I also recommend putting handler-specific settings in a handler-specific options type.

To access the enum type then we would have to add the conditional compilation guard around it. See the conditional compilation guards around the `SocketHttpHandler` code for reference

src/Momento.Sdk/Config/Transport/IGrpcConfiguration.cs
malandis: Agreed. This could be confusing otherwise if someone is in a runtime that doesn't use `SocketsHttpHandler` but then they attempt to override this setting thinking it applies unconditionally.

src/Momento.Sdk/Config/Transport/StaticTransportStrategy.cs
```diff
@@ -16,9 +17,15 @@ public class StaticGrpcConfiguration : IGrpcConfiguration
     /// <inheritdoc/>
     public int MinNumGrpcChannels { get; }
     /// <inheritdoc/>
-    public GrpcChannelOptions GrpcChannelOptions { get; }
+    public GrpcChannelOptions? GrpcChannelOptions { get; }
```
malandis, line 20: This line triggered warnings in GitHub actions. Would you look at that? Otherwise looks good to go 👍

malandis (approved):

PR #539, opened 2024-02-28
feat: adds customization options for SocketsHttpHandler

malandis:
# Overview
1. Add SocketHttpHandler config options and plumb through.
2. Set Lambda config to use `SocketHttpHandler.PooledConnectionIdleTimeout` 
    of 6 minutes. 
3. Add target frameworks for .NET 6.0 and bump target framework net461
    to net462 for async iterator support.

# SocketHttpHandler config
In `Grpc.Net.Client`, the default HttpHandler on .NET runtimes .NET
5.0 or greater is `SocketsHttpHandler`. In this PR we create a new
options class to encapsulate customizations to the handler, most
importantly the pooled connection timeout and whether to enable
multiple http connections.

# Lambda config
The default value for the idle connection timeout is 1 minute. We find
this too strict in lambda environments, where the container may freeze
and thaw in greater than 1 minute spans but less than the server
timeout (5 minutes as of writing). Therefore we set the lambda config
pooled connection idle timeout higher, to 6 minutes. That way a lambda
function will not needlessly reconnect when the connection is still
open.

# Target framework
Previously we relied on .NET Framework 4.61 as the .NET Framework
target. The next minor version introduced async iterator features
necessary for the topic client. Because of this, at the time, we added
preprocessor directives to exclude the topic client if the build target
wasn't .NET Standard 2.0 or greater.

Now that we add another target framework, .NET 6.0, maintaining the
different platform directives is difficult. Instead we bump the Framework
target from net461 to net462 to gain the async iterators and remove
the platform conditional compilation directives. There is no customer
impact by removing exactly .NET 4.61 as a target framework.

.NET 4.62 also includes `WinHttpHandler`, so we can revisit the use of
gRPC web in that Framework.

### DIFF
--- a/src/Momento.Sdk/Config/Transport/SocketsHttpHandlerOptions.cs
+++ b/src/Momento.Sdk/Config/Transport/SocketsHttpHandlerOptions.cs
@@ -1,0 +1,66 @@
+#pragma warning disable 1591
+using System;
+using Momento.Sdk.Internal;
+namespace Momento.Sdk.Config.Transport;
+
+public class SocketsHttpHandlerOptions
+{
+    public static TimeSpan DefaultPooledConnectionIdleTimeout { get; } = TimeSpan.FromMinutes(1);
+    public TimeSpan PooledConnectionIdleTimeout { get; } = DefaultPooledConnectionIdleTimeout;
+    public bool EnableMultipleHttp2Connections { get; } = true;
+
+    public SocketsHttpHandlerOptions() { }
+    public SocketsHttpHandlerOptions(TimeSpan pooledConnectionIdleTimeout) : this(pooledConnectionIdleTimeout, true) { }
+    public SocketsHttpHandlerOptions(bool enableMultipleHttp2Connections) : this(DefaultPooledConnectionIdleTimeout, enableMultipleHttp2Connections) { }
+
+    public SocketsHttpHandlerOptions(TimeSpan pooledConnectionIdleTimeout, bool enableMultipleHttp2Connections)
+    {
+        Utils.ArgumentStrictlyPositive(pooledConnectionIdleTimeout, nameof(pooledConnectionIdleTimeout));
+        PooledConnectionIdleTimeout = pooledConnectionIdleTimeout;
+        EnableMultipleHttp2Connections = enableMultipleHttp2Connections;
+    }
+
+    public SocketsHttpHandlerOptions WithPooledConnectionIdleTimeout(TimeSpan pooledConnectionIdleTimeout)
+    {
+        return new SocketsHttpHandlerOptions(pooledConnectionIdleTimeout, EnableMultipleHttp2Connections);
+    }
+
+    public SocketsHttpHandlerOptions WithEnableMultipleHttp2Connections(bool enableMultipleHttp2Connections)
+    {
+        return new SocketsHttpHandlerOptions(PooledConnectionIdleTimeout, enableMultipleHttp2Connections);
+    }
+
+    public static SocketsHttpHandlerOptions Of(TimeSpan pooledConnectionIdleTimeout)
+    {
+        return new SocketsHttpHandlerOptions(pooledConnectionIdleTimeout);
+    }
+
+    public static SocketsHttpHandlerOptions Of(bool enableMultipleHttp2Connections)
+    {
+        return new SocketsHttpHandlerOptions(enableMultipleHttp2Connections);
+    }
+
+    public static SocketsHttpHandlerOptions Of(TimeSpan pooledConnectionIdleTimeout, bool enableMultipleHttp2Connections)
+    {
+        return new SocketsHttpHandlerOptions(pooledConnectionIdleTimeout, enableMultipleHttp2Connections);
+    }
+
+    public override bool Equals(object obj)
+    {
+        if (obj == null || GetType() != obj.GetType())
+        {
+            return false;
+        }
+
+        var other = (SocketsHttpHandlerOptions)obj;
+        return PooledConnectionIdleTimeout.Equals(other.PooledConnectionIdleTimeout) &&
+            EnableMultipleHttp2Connections.Equals(other.EnableMultipleHttp2Connections);
+    }
+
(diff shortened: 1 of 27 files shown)


src/Momento.Sdk/Internal/ControlGrpcManager.cs
```diff
@@ -98,20 +100,24 @@ public ControlGrpcManager(ILoggerFactory loggerFactory, string authToken, string
...
             MaxReceiveMessageSize = Internal.Utils.DEFAULT_MAX_MESSAGE_SIZE,
             MaxSendMessageSize = Internal.Utils.DEFAULT_MAX_MESSAGE_SIZE,
-#if USE_GRPC_WEB
-            HttpHandler = new GrpcWebHandler(new HttpClientHandler()),
+#if NET5_0_OR_GREATER
+            HttpHandler = new System.Net.Http.SocketsHttpHandler
+            {
+                EnableMultipleHttp2Connections = config.TransportStrategy.GrpcConfig.SocketsHttpHandlerOptions.EnableMultipleHttp2Connections,
+                PooledConnectionIdleTimeout = config.TransportStrategy.GrpcConfig.SocketsHttpHandlerOptions.PooledConnectionIdleTimeout
+            }
```
malandis, line 108: Here is where we set the SocketsHttpHandler settings for the control client

src/Momento.Sdk/Internal/DataGrpcManager.cs
```diff
@@ -282,16 +283,20 @@ internal DataGrpcManager(IConfiguration config, string authToken, string endpoin
...
         channelOptions.MaxSendMessageSize = Internal.Utils.DEFAULT_MAX_MESSAGE_SIZE;
-        
-#if USE_GRPC_WEB
+
+#if NET5_0_OR_GREATER
+        channelOptions.HttpHandler = new SocketsHttpHandler
+        {
+            EnableMultipleHttp2Connections = config.TransportStrategy.GrpcConfig.SocketsHttpHandlerOptions.EnableMultipleHttp2Connections,
+            PooledConnectionIdleTimeout = config.TransportStrategy.GrpcConfig.SocketsHttpHandlerOptions.PooledConnectionIdleTimeout
+        };
```
malandis, line 292: Here is where we set the data client SocketsHttpHandler settings

src/Momento.Sdk/Internal/LoggingUtils.cs
```diff
@@ -379,8 +379,6 @@ public static TSuccess LogTraceCollectionRequestSuccess<TSuccess>(this ILogger _
             }
             return success;
         }
-        
-#if NETSTANDARD2_0_OR_GREATER
```
malandis, line 383: None of these NETSTANDARD2.0 directives are necessary anymore since I bumped the target framework from .NET 4.61 to .NET 4.62.

src/Momento.Sdk/Momento.Sdk.csproj
```diff
@@ -2,11 +2,9 @@
 <Project Sdk="Microsoft.NET.Sdk">
 	<PropertyGroup>
 		<!-- Build Configuration -->
-		<TargetFrameworks>netstandard2.0;net461</TargetFrameworks>
+		<TargetFrameworks>netstandard2.0;net6.0;net462</TargetFrameworks>
```
malandis, line 5: Adding net6.0 and bumping net 4.61 to 4.62.

src/Momento.Sdk/Momento.Sdk.csproj
```diff
@@ -2,11 +2,9 @@
...
 		<LangVersion>latest</LangVersion>
 		<Nullable>enable</Nullable>
-		<!-- Microsoft.Bcl.AsyncInterfaces on net461 isn't supported; suppress warning -->
-		<SuppressTfmSupportBuildWarnings>true</SuppressTfmSupportBuildWarnings>
```
malandis, line 9: No longer necessary since we are on net462

src/Momento.Sdk/Config/Transport/SocketsHttpHandlerOptions.cs
```diff
@@ -0,0 +1,66 @@
+#pragma warning disable 1591
+using System;
+using Momento.Sdk.Internal;
+namespace Momento.Sdk.Config.Transport;
+
+public class SocketsHttpHandlerOptions
+{
+    public static TimeSpan DefaultPooledConnectionIdleTimeout { get; } = TimeSpan.FromMinutes(1);
+    public TimeSpan PooledConnectionIdleTimeout { get; } = DefaultPooledConnectionIdleTimeout;
+    public bool EnableMultipleHttp2Connections { get; } = true;
```
pratik151192, line 10: Curious if we needed in this PR and if `true` is default (should be but confirming). I am having a hard time connect this with gRPC as ideally the http2 control should be in the gRPC libraries.

malandis (commented):
Left some clarifying comments. Lots of warnings leftover that we need to review.

src/Momento.Sdk/Internal/DataGrpcManager.cs
```diff
@@ -282,16 +283,20 @@ internal DataGrpcManager(IConfiguration config, string authToken, string endpoin
...
+#elif USE_GRPC_WEB
         channelOptions.HttpHandler = new GrpcWebHandler(new HttpClientHandler());
```
pratik151192, line 294: Okay nice since the grpc handler takes this it shouldn't affect global state !

src/Momento.Sdk/Config/Transport/SocketsHttpHandlerOptions.cs
malandis: This is to align with the gRPC .NET client defaults. We can revisit later: https://github.com/grpc/grpc-dotnet/blob/098dca892a3949ade411c3f2f66003f7b330dfd2/src/Shared/HttpHandlerFactory.cs#L27-L51

pratik151192 (approved):

nand4011 (approved):

PR #542, opened 2024-03-05
fix: disable keepalive for control clients

anitarua:


src/Momento.Sdk/Internal/VectorIndexControlClient.cs
```diff
@@ -22,7 +23,17 @@ internal sealed class VectorIndexControlClient : IDisposable
...
     public VectorIndexControlClient(IVectorIndexConfiguration config, string authToken, string endpoint)
     {
-        grpcManager = new VectorIndexControlGrpcManager(config, authToken, endpoint);
+        // Override the sockets http handler options to disable keepalive
+        var overrideKeepalive = SocketsHttpHandlerOptions.Of(
+            pooledConnectionIdleTimeout: config.TransportStrategy.GrpcConfig.SocketsHttpHandlerOptions.PooledConnectionIdleTimeout,
+            enableMultipleHttp2Connections: config.TransportStrategy.GrpcConfig.SocketsHttpHandlerOptions.EnableMultipleHttp2Connections,
+            keepAlivePingTimeout: System.Threading.Timeout.InfiniteTimeSpan,
+            keepAlivePingDelay: System.Threading.Timeout.InfiniteTimeSpan,
+            keepAlivePermitWithoutCalls: false
```
malandis, line 32: What do these settings default to in the data client?

anitarua: [Defaults](https://github.com/momentohq/client-sdk-dotnet/blob/4db09f872e8e9a67320b5e4f03bb92d7840b4dc3/src/Momento.Sdk/Config/Transport/SocketsHttpHandlerOptions.cs#L21-L41) to enabling keepalives with:
* timeout: 1000ms
* delay: 5000ms
* permit: true

malandis: OK. Leaving this here for the future: previously we did not have keepalives enabled at all in .NET since `SocketsHttpHandler` defaults to no keepalive (unlike other gRPC libraries?). So if an existing user upgrades the settings will be different, though these settings have the most impact in Lambda environments, which we've already tuned.

malandis (approved):

PR #543, opened 2024-03-06
feat: drop grpc-web for .NET 4.62 build.

malandis:
Since we bumped the minimum .NET Framework build version from 4.61 to
4.62, we gained access to an Http/2 handler for .NET
Framework. Because 4.61 did not have an Http/2 handler, we had to
resort to gRPC-Web. Now we only need gRPC-Web for the Unity build.

In order to exercise the gRPC-Web build, we add that to the testing
matrix.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -14,6 +14,7 @@ jobs:
         include:
           - os: windows-latest
             target-framework: net462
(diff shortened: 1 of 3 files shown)


bruuuuuuuce (approved):
:shipit:

PR #544, opened 2024-03-06
chore: update readmes to use api key from console instead of auth token from cli

anitarua:
Clean up readme instructions from "auth token from CLI" to "api key from console"


malandis (approved):
Thanks!

PR #546, opened 2024-03-06
fix: generate readme should be its own ci job

anitarua:
Noticed github actions failures recently when updating the readme and Matt pointed out the readme generation step was being run twice, once for the windows build and once for the ubuntu build, causing a race condition and causing one of them to fail because the readme had already been generated.

This PR extracts the readme generation into a separate github actions job in the same workflow file so that will run only once.


.github/workflows/on-push-to-main-branch.yaml
```diff
@@ -54,4 +57,4 @@ jobs:
           sdk_language: .NET
           template_file: ./README.template.md
           output_file: ./README.md
-          dev_docs_slug: dotnet
+          dev_docs_slug: dotnet
```
bruuuuuuuce, line 60: nit: can we add newline here so the 🛑 thing goes away 😄

bruuuuuuuce (dismissed):
:shipit:

bruuuuuuuce (approved):
:shipit:

PR #549, opened 2024-03-15
fix: dispose cache client when eager connection fails

malandis:
In order to clean up resources on a failed eager connection (`Ping`)
RPC, we dispose the cache client on exception in
`CacheClient::CreateAsync`.


bruuuuuuuce (approved):
:shipit:

PR #550, opened 2024-03-15
fix: properly re-throw caught exception in eager connection code

malandis:
To preserve the original stack trace, one needs to use a `throw;`
statement, not re-throw the original exception. See https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/statements/exception-handling-statements


nand4011 (dismissed):
LGTM, but github really doesn't like that e isn't used.

malandis:
> LGTM, but github really doesn't like that e isn't used.

Fixed in [307cf63](https://github.com/momentohq/client-sdk-dotnet/pull/550/commits/307cf635874f47b14fca352cb78dce0bf0f5a3f4)

nand4011 (approved):

PR #553, opened 2024-04-18
chore: rename TEST_AUTH_TOKEN to MOMENTO_API_KEY

pgautier404:


.github/workflows/build.yaml
```diff
@@ -20,7 +20,7 @@ jobs:
             grpc-web: true
     runs-on: ${{ matrix.os }}
     env:
-      TEST_AUTH_TOKEN: ${{ secrets.ALPHA_TEST_AUTH_TOKEN }}
+      TEST_API_KEY: ${{ secrets.ALPHA_TEST_AUTH_TOKEN }}
```
cprice404, line 23: thanks for changing this - but if we're going to rename it I'd honestly rather we just rename it to MOMENTO_API_KEY. I have never really understood why we are differentiating between test vs. prod when choosing the env var name. In the future I am imagining that a lot of the SDKs will support a default credential provider chain where you don't have to pass a cred provider if you don't want, and if you don't, it will look in some default places (env var, ~/.momento files, etc.). In that future I am expecting the env var name to be MOMENTO_API_KEY so I'd prefer to just be working towards using that env var name everywhere.

cprice404 (approved):
dance

PR #554, opened 2024-04-26
spike: demo the storage api in C#

malandis:
This demonstrates the Store API in C#. Because C# has significant support for implicit casts, we can greatly simplify the developer experience when interacting with the multitude of types that go in and out of the store.

By using implicit casts on the way in, we only need one `Put` method that takes a `StoreValue`. The `StoreValue` type defines casts from various types to `StoreValue`. That way a developer can write:

```dotnet
await client.PutAsync("my-store", "my-key", 3.14);
```

and the runtime will create a `StoreValue` from the double. The compiler detects any invalid casts.

Similarly on when we retrieve data with `GetAsync`, we can return a `Success` that has a `StoreValue`. The user can benefit from implicit casting in the reverse direction. They can write

```dotnet
var getResponse = await client.GetAsync("my-store", "string-value");
if (getResponse is StoreGet.Success success)
{
    string value = success.Value;
}
```

If the cast is incorrect, the .NET runtime will throw an `InvalidCastException`.

By storing the type in `StoreValue`, we the developer can do pattern matching exhaustively over that. The compiler notifies the developer when the pattern matching is not exhaustive.

### DIFF
--- a/examples/StoreApi/Program.cs
+++ b/examples/StoreApi/Program.cs
@@ -1,0 +1,85 @@
+﻿using Momento.Sdk.StoreClient;
+
+namespace Driver;
+
+public class Program
+{
+    public static async Task Main(string[] args)
+    {
+        var client = new StoreClient();
+        await client.CreateStoreAsync("my-store");
+
+        // Can use implicit cast to create StoreValue.
+        await client.PutAsync("my-store", "my-key", "my-value");
+        await client.PutAsync("my-store", "my-key", 42);
+        await client.PutAsync("my-store", "my-key", 3.14);
+        await client.PutAsync("my-store", "my-key", true);
+
+        // The compiler will detect invalid casts.
+        // await client.PutAsync("my-store", "my-key", new string[] { "a", "b" });
+
+        // This demos usage where the developer is sure of the type.
+        {
+            var response = await client.GetAsync("my-store", "string-value");
+            if (response is StoreGet.Success success)
+            {
+                // This uses an implicit cast from StoreValue to string.
+                string value = success.Value;
+                Console.WriteLine($"Success: {value}");
+            }
+            else if (response is StoreGet.Error error)
+            {
+                Console.WriteLine($"Error: {error.Message}");
+            }
+        }
+
+        // A more cautious developer would handle the unfortunate case where the cast fails.
+        {
+            var response = await client.GetAsync("my-store", "bool-value");
+            if (response is StoreGet.Success success)
+            {
+                try
+                {
+                    // This uses an implicit cast from StoreValue to long, but it's actually a bool.
+                    long value = success.Value;
+                    Console.WriteLine($"Success: {value}");
+                }
+                catch (InvalidCastException)
+                {
+                    Console.WriteLine($"Error: Invalid cast to long. Type was actually {success.Value.Type}");
+                }
+            }
+            else if (response is StoreGet.Error error)
+            {
+                Console.WriteLine($"Error: {error.Message}");
+            }
+        }
+
+        // This demos usage where the developer wants to do something specific based on the type.
+        {
+            foreach (var key in new string[] { "string-value", "int-value", "double-value", "bool-value", "unknown-value" })
+            {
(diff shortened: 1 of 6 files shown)


examples/StoreApi/StoreValue.cs
```diff
@@ -0,0 +1,49 @@
...
+    public StoreValue(double value) : this(value, StoreValueType.Double)
+    {
+    }
+
+    public StoreValue(bool value) : this(value, StoreValueType.Bool)
+    {
+    }
+
+    // implicit cast from the various types into StoreValue
+    public static implicit operator StoreValue(string value) => new StoreValue(value);
```
cprice404, line 39: theoretically the right-hand-side of these could be a code block, correct? Allowing us to put in some logging or throw a different error type if we decide to, etc.?

Not saying we should change it from what you have here, just wanted to understand our options.

examples/StoreApi/Program.cs
```diff
@@ -0,0 +1,85 @@
...
+    public static async Task Main(string[] args)
+    {
+        var client = new StoreClient();
+        await client.CreateStoreAsync("my-store");
+
+        // Can use implicit cast to create StoreValue.
+        await client.PutAsync("my-store", "my-key", "my-value");
+        await client.PutAsync("my-store", "my-key", 42);
+        await client.PutAsync("my-store", "my-key", 3.14);
+        await client.PutAsync("my-store", "my-key", true);
```
cprice404, line 16: kewl

examples/StoreApi/Program.cs
```diff
@@ -0,0 +1,85 @@
...
+            var response = await client.GetAsync("my-store", "bool-value");
+            if (response is StoreGet.Success success)
+            {
+                try
+                {
+                    // This uses an implicit cast from StoreValue to long, but it's actually a bool.
+                    long value = success.Value;
+                    Console.WriteLine($"Success: {value}");
+                }
+                catch (InvalidCastException)
```
cprice404, line 47: This one makes me want to raise the same question I raised on your JS PR about whether we should be modeling Success/Error as algebraic types, or just return the value directly and throw an exception for errors. In this case, that would simplify the code down to a try/catch block with multiple catch clauses, as opposed to an if+try.

Probably it'd be too inconsistent to make that dramatic of a change for this Client given that it will be living in the same SDK, but I'm just trying to think through what I would advocate for here if it was a greenfield product.

examples/StoreApi/StoreValue.cs
malandis: Yes. The right-hand side here is just a shorthand when the function body is a single statement. We can have a multi-statement block instead.

examples/StoreApi/Program.cs
malandis: Same comment re: the JS API. I like the shorthand but there's forward compatibility concerns.

cprice404: In what sense? Are you talking about the case where the server changes in the future to support additional response types?

malandis: Nevermind -- I was confusing the inner value type with the `Success` object. You're saying we can return the `Success` object or bust. We are free to add more data to `Success` if we do that.

examples/StoreApi/Program.cs
```diff
@@ -0,0 +1,85 @@
...
+        await client.PutAsync("my-store", "my-key", 3.14);
+        await client.PutAsync("my-store", "my-key", true);
+
+        // The compiler will detect invalid casts.
+        // await client.PutAsync("my-store", "my-key", new string[] { "a", "b" });
+
+        // This demos usage where the developer is sure of the type.
+        {
+            var response = await client.GetAsync("my-store", "string-value");
+            if (response is StoreGet.Success success)
```
cprice404, line 24: have you seen this before?

https://spencerfarley.com/2021/03/26/unions-in-csharp/#improvements-in-c

malandis: It rings a bell. I think C# 9 wasn't out at the time and was a preview.

PR #556, opened 2024-05-24
feat: bump grpc.net.client to v2.63.0

malandis:
This addresses regressions we saw when spawning many new
connections. Possibly fixed by https://github.com/grpc/grpc-dotnet/pull/2422


cprice404 (approved):
increase

PR #562, opened 2024-06-27
chore: lowercase agent and runtime-headers

malandis:
In order to standardize the agent and runtime-version headers, we
lowercase the header names.

### DIFF
--- a/src/Momento.Sdk/Internal/Middleware/HeaderMiddleware.cs
+++ b/src/Momento.Sdk/Internal/Middleware/HeaderMiddleware.cs
@@ -11,9 +11,9 @@ namespace Momento.Sdk.Internal.Middleware
 {
     internal class Header
     {
-        public const string AuthorizationKey = "Authorization";
-        public const string AgentKey = "Agent";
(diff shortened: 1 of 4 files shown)


bruuuuuuuce (approved):
:shipit:

src/Momento.Sdk/Internal/Middleware/HeaderMiddleware.cs
```diff
@@ -11,9 +11,9 @@ namespace Momento.Sdk.Internal.Middleware
 {
     internal class Header
     {
-        public const string AuthorizationKey = "Authorization";
-        public const string AgentKey = "Agent";
-        public const string RuntimeVersionKey = "Runtime_Version";
```
malandis, line 16: @bruuuuuuuce any idea why this was underscored?

bruuuuuuuce: was either old or must have been a typo, we have aligned with backend team that we want to use `runtime-version`

PR #567, opened 2024-08-14
refactor: organize targets by service not endpoint

malandis:
Previously we organized the make targets by endpoint. This isn't
granular enough to test when some services aren't deployed at an
endpoint (eg leaderboards in the cache endpoint).

With this change we test by service. Because of this we simplify the
test organization by not grouping by control vs data.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -8,7 +8,7 @@
 #   - On other operating systems `make build` (test) we only runs the .NET 6.0 build (test) targets.
 # - We also have a GRPC_WEB flag that can be set to true to enable gRPC-Web support.
 #   - The caller can run `make GRPC_WEB=true build` to enable gRPC-Web support.
-# - We additionally group the integration tests by service (cache, control, token).
+# - We additionally group the integration tests by service (auth, cache, topics).
 #   - This is to allow for more granular testing by service.
 #   - Similar to `build` and `test` targets, we have `test-cache-service`, `test-topics-service`, `test-auth-service` targets
 #	  that are conditionally run based on the operating system.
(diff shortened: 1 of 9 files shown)


Makefile
```diff
@@ -32,15 +32,15 @@ TEST_LOGGER_OPTIONS := --logger "console;verbosity=detailed"
 ifneq (,$(findstring NT,$(OS)))
 	BUILD_TARGETS := build-dotnet6 build-dotnet-framework
 	TEST_TARGETS := test-dotnet6 test-dotnet-framework
-	TEST_TARGETS_CACHE_ENDPOINT := test-dotnet6-cache-endpoint test-dotnet-framework-cache-endpoint
-	TEST_TARGETS_CONTROL_ENDPOINT := test-dotnet6-control-endpoint test-dotnet-framework-control-endpoint
-	TEST_TARGETS_TOKEN_ENDPOINT := test-dotnet6-token-endpoint test-dotnet-framework-token-endpoint
+	TEST_TARGETS_AUTH_SERVICE := test-dotnet6-auth-service test-dotnet-framework-auth-service
+	TEST_TARGETS_CACHE_SERVICE := test-dotnet6-cache-service test-dotnet-framework-cache-service
```
cprice404, line 36: @malandis I think we still need to keep control plane tests separate since it has a different endpoint, right? sorry if I missed calling this out when you pinged me in slack the other day. lmk if you want to zoom about it

malandis: Thanks for clearing that up. As discussed we can loop back and split this up additionally by endpoint.

cprice404 (approved):
fine to merge this and then come back and hit the endpoint stuff

PR #573, opened 2024-09-19
fix: add a sleep in a list test to fix an intermittent canary failure

nand4011:


tests/Integration/Momento.Sdk.Tests/Cache/ListTest.cs
```diff
@@ -727,7 +727,9 @@ public async Task ListPushFrontFetch_ValueIsString_HappyPath()
         Assert.True(pushResponse is CacheListPushFrontResponse.Success, $"Unexpected response: {pushResponse}");
         successResponse = (CacheListPushFrontResponse.Success)pushResponse;
         Assert.Equal(2, successResponse.ListLength);
-
+        
+        Thread.Sleep(1000);
```
cprice404, line 731: is the theory here that the subsequent fetch is hitting a replica? if so, do we not want to consider the same consistent read story? or is this just intended to make a short term improvement while we get more data about the consistent reads?

PR #575, opened 2024-09-19
feat: add an topic event enumerable

malandis:
Introduces an interface `ITopicEvent` which covers both normal
 messages (`TopicMessage`) and system events (`TopicSystemEvent`). We
refactor the internals of the stream to be over `ITopicEvent`s, so
that we can filter to just `TopicMessage` (as is implemented
currently) or pass through all events in an event enumerable.

To enumerate over all events in the stream, call `GetAllEventsAsyncEnumerable`,
or `WithCancellationForAllEvents`.

This largely parallels the existing enumerator over messages only. See the
integration tests for an example usage.

### DIFF
--- a/src/Momento.Sdk/Internal/ScsTopicClient.cs
+++ b/src/Momento.Sdk/Internal/ScsTopicClient.cs
@@ -198,7 +198,7 @@ internal sealed class ScsTopicClient : ScsTopicClientBase
             _subscribed = true;
         }
 
-        public async ValueTask<TopicMessage?> GetNextRelevantMessageFromGrpcStreamAsync(
+        public async ValueTask<ITopicEvent?> GetNextRelevantMessageFromGrpcStreamAsync(
             CancellationToken cancellationToken)
         {
             while (!cancellationToken.IsCancellationRequested)
--- a/src/Momento.Sdk/Internal/ScsTopicClient.cs
+++ b/src/Momento.Sdk/Internal/ScsTopicClient.cs
@@ -142,7 +142,7 @@ internal sealed class ScsTopicClient : ScsTopicClientBase
         }
 
         var response = new TopicSubscribeResponse.Subscription(
-            cancellationToken => subscriptionWrapper.GetNextRelevantMessageFromGrpcStreamAsync(cancellationToken),
+            cancellationToken => subscriptionWrapper.GetNextEventFromGrpcStreamAsync(cancellationToken),
             subscriptionWrapper.Dispose);
         return _logger.LogTraceTopicRequestSuccess(RequestTypeTopicSubscribe, cacheName, topicName,
             response);
@@ -198,7 +198,7 @@ internal sealed class ScsTopicClient : ScsTopicClientBase
             _subscribed = true;
         }
 
-        public async ValueTask<ITopicEvent?> GetNextRelevantMessageFromGrpcStreamAsync(
+        public async ValueTask<ITopicEvent?> GetNextEventFromGrpcStreamAsync(
             CancellationToken cancellationToken)
         {
             while (!cancellationToken.IsCancellationRequested)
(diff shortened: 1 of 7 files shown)


malandis:
Opening for early review

nand4011:
'event' feels like 'a thing that happened' to me. I like it for `TopicSystemEvent`, but I'm not as sure about `ITopicEvent` also representing normal messages. Maybe `ITopicItem`? I don't have too strong an opinion on it though.
The same thing with `GetEventAsyncEnumerable`. By the name alone I would assume it would be for exclusively system events, but the docs would clear that up. `GetUnabridgedAsyncEnumerable` could work, but might be too much of a mouthful.

src/Momento.Sdk/Internal/ScsTopicClient.cs
```diff
@@ -198,7 +198,7 @@ public async Task Subscribe()
             _subscribed = true;
         }
 
-        public async ValueTask<TopicMessage?> GetNextRelevantMessageFromGrpcStreamAsync(
+        public async ValueTask<ITopicEvent?> GetNextRelevantMessageFromGrpcStreamAsync(
```
nand4011, line 201: If we change this to return system events as well, we could shorten my Java-length method name choice here.

cprice404:
> 'event' feels like 'a thing that happened' to me. I like it for `TopicSystemEvent`, but I'm not as sure about `ITopicEvent` also representing normal messages. Maybe `ITopicItem`? I don't have too strong an opinion on it though. The same thing with `GetEventAsyncEnumerable`. By the name alone I would assume it would be for exclusively system events, but the docs would clear that up. `GetUnabridgedAsyncEnumerable` could work, but might be too much of a mouthful.

There was a thread in slack when we did this work for golang, and Event is the name that most people settled on, so as far as the public API for this goes I think we should be trying to mirror (semantically) what we did in go. I think it probably makes sense for the parent type / interface to include the word Event even though I do agree there is a bit of ambiguity for normal messages.

cprice404 (commented):
i think i'm good with the names so far.

might be good to get @anitarua to weigh in since she did the prev two.

anitarua:
Yeah, I think the names are fine here, `ITopicEvent` implemented by `TopicMessage` and `TopicSystemEvent`.

In the Go SDK, `TopicEvent` was implemented by `TopicItem` (string or byte messages), `TopicHeartbeat`, and `TopicDiscontinuity`.

src/Momento.Sdk/Internal/ScsTopicClient.cs
malandis: Done in https://github.com/momentohq/client-sdk-dotnet/pull/575/commits/9d8a4bb99e29aee72b0806c829ffb2818d722520

malandis:
Given the consensus, I've left the names as is. As Anita pointed out, there's a difference in naming vs the Golang SDK.

In the protos we have `SubscriptionItem` and `TopicItem`. `SubscriptionItem` now matches `ITopicEvent` and covers text, binary, heartbeat, and discontinuities. `TopicItem` in the protos has been called `TopicMessage` in .NET. We'll leave the latter as is for backwards compatibility.

tests/Integration/Momento.Sdk.Tests/Topics/TopicTest.cs
```diff
@@ -143,10 +143,67 @@ public async Task PublishAndSubscribe_String_Succeeds()
...
+        {
+            "one",
+            "two",
+            "three",
+            "four",
+            "five"
+        };
+
+        var produceCancellation = new CancellationTokenSource();
+        //produceCancellation.CancelAfter(2000);
```
nand4011, line 166: nit: looks like this can be removed.

nand4011 (approved):

tests/Integration/Momento.Sdk.Tests/Topics/TopicTest.cs
malandis: Resolved in [dd6ab9d](https://github.com/momentohq/client-sdk-dotnet/pull/575/commits/dd6ab9dfc7a23d868ec5a4c4aad8939bf6cfa562)

nand4011 (approved):

PR #578, opened 2024-09-24
chore: add example logging if heartbeat not received

malandis:
Adds an example demonstrating logging a message if a heartbeat isn't
received in an interval of time.


cprice404 (approved):
✨

PR #581, opened 2024-09-25
feat: Add the read concern header

nand4011:
Add the read concern header to the top level Configuration.

Set the consistent read concern header in the cache client test fixture if the CONSISTENT_READS env var is set.

Add a new make target for the github build to use that sets the header.

Update the test-cache-service make target to set the header so that the canaries will use consistent reads.


src/Momento.Sdk/Internal/DataGrpcManager.cs
```diff
@@ -256,10 +250,13 @@ public class DataGrpcManager : GrpcManager
 
     internal DataGrpcManager(IConfiguration config, string authToken, string endpoint): base(config.TransportStrategy.GrpcConfig, config.LoggerFactory, authToken, endpoint, "DataGrpcManager")
     {
+        var readConcernHeader = new Header(Header.ReadConcern, config.ReadConcern.ToStringValue());
+        headers.Add(readConcernHeader);
```
nand4011, line 254: Do we need to add the read concern header if it's set to balanced? It seems like extra bytes over the wire for no difference. I suppose the default may change in the future.

nand4011:
The tests all pass locally for me with consistent reads turned on, so I would be alright with removing the separate make targets if someone else can verify that.

src/Momento.Sdk/Config/Configuration.cs
```diff
@@ -34,31 +56,34 @@ public Configuration(ILoggerFactory loggerFactory, IRetryStrategy retryStrategy,
     /// <param name="retryStrategy">Defines a contract for how and when to retry a request</param>
     /// <param name="middlewares">The Middleware interface allows the Configuration to provide a higher-order function that wraps all requests.</param>
     /// <param name="transportStrategy">This is responsible for configuring network tunables.</param>
-    /// <param name="loggerFactory">This is responsible for configuraing logging.</param>
-    public Configuration(ILoggerFactory loggerFactory, IRetryStrategy retryStrategy, IList<IMiddleware> middlewares, ITransportStrategy transportStrategy)
+    /// <param name="loggerFactory">This is responsible for configuring logging.</param>
+    /// <param name="readConcern">The client-wide setting for read-after-write consistency.</param>
+    public Configuration(ILoggerFactory loggerFactory, IRetryStrategy retryStrategy, IList<IMiddleware> middlewares,
+        ITransportStrategy transportStrategy, ReadConcern readConcern)
```
malandis, line 62: You can set a default value to read concern `ReadConcern readConcern = ReadConcern.Balanced` I believe. Then no need for the overload.

src/Momento.Sdk/Config/ReadConcern.cs
```diff
@@ -0,0 +1,41 @@
...
+public enum ReadConcern
+{
+    /// <summary>
+    /// Balanced is the default read concern for the cache client.
+    /// </summary>
+    Balanced,
+    /// <summary>
+    /// Consistent read concern guarantees read after write consistency.
+    /// </summary>
+    Consistent
```
malandis, line 18: Isn't there a third value?

src/Momento.Sdk/Internal/DataGrpcManager.cs
cprice404: i agree with your comment @nand4011 , I would omit it unless it's explicitly specified in the config

cprice404 (commented):
agree with the comment about not setting the header unless it's explicitly specified in the config

src/Momento.Sdk/Config/Configuration.cs
nand4011: I was trying to think if having two constructors would be better than a default argument for backwards compatibility if we add more arguments in the future. I'm going to make it nullable anyway and have it not explicitly set by default.

src/Momento.Sdk/Config/ReadConcern.cs
nand4011: Not in the Go or JS SDKs

src/Momento.Sdk/Config/Configuration.cs
malandis: We should just go with a default value. Otherwise we would have to document what is going to happen when it is null, which will be we set to the default value.

malandis (approved):

PR #584, opened 2024-10-31
chore: improve resource exhausted error message

anitarua:
Closes https://github.com/momentohq/dev-eco-issue-tracker/issues/1039

Passes the `RpcException` to the `LimitExceededError` so it can parse the trailer metadata or error details to determine the most specific error message wrapper to return.

### DIFF
--- a/src/Momento.Sdk/Exceptions/LimitExceededException.cs
+++ b/src/Momento.Sdk/Exceptions/LimitExceededException.cs
@@ -14,31 +14,31 @@ public class LimitExceededException : SdkException
         if (errMetadata != null) {
             this.MessageWrapper = errMetadata switch
             {
-                "topic_subscriptions_limit_exceeded" => LimitExceededMessageWrapper.TopicSubscriptionsLimitExceeded.Value,
-                "operations_rate_limit_exceeded" => LimitExceededMessageWrapper.OperationsRateLimitExceeded.Value,
-                "throughput_rate_limit_exceeded" => LimitExceededMessageWrapper.ThroughputRateLimitExceeded.Value,
-                "request_size_limit_exceeded" => LimitExceededMessageWrapper.RequestSizeLimitExceeded.Value,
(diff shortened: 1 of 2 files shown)


src/Momento.Sdk/Exceptions/LimitExceededException.cs
```diff
@@ -1,15 +1,91 @@
...
+    private LimitExceededMessageWrapper(string value)
+    {
+        Value = value;
+    }
+
+    /// <summary>
+    /// The value of the LimitExceededMessageWrapper.
+    /// </summary>
+    public string Value { get; private set; }
+}
```
malandis, line 91: I like centralizing the strings in one place.

We can simplify this: this can be a static class `LimitExceededMessages` with static string fields, ie no need for a type to capture the value. Unless you see a need for additional fields or behavior?

src/Momento.Sdk/Exceptions/LimitExceededException.cs
```diff
@@ -1,15 +1,91 @@
...
+            } else if (lowerCasedMessage.Contains("item size")) {
+                this.MessageWrapper = LimitExceededMessageWrapper.ItemSizeLimitExceeded.Value;
+            } else if (lowerCasedMessage.Contains("element size")) {
+                this.MessageWrapper = LimitExceededMessageWrapper.ElementSizeLimitExceeded.Value;
+            } else {
+                this.MessageWrapper = LimitExceededMessageWrapper.UnknownLimitExceeded.Value;
+            }
+        }
     }
 }
```
malandis, line 45: For separation of concerns, we let the constructor initialize the instance and separate out resolving the `MessageWrapper`. We can make a static method `CreateWithMessageWrapper` (or name of your choice) that resolves the `MessageWrapper`, then returns an instance with that set (by having a constructor that takes the `MessageWrapper`.

malandis (approved):
Lgtm!

PR #589, opened 2024-11-12
feat: use ulong and topics and plumb through `resumeAt` subscription arguments

malandis:
Changes TopicItem and TopicSystemEvent to store the sequence number/page as `ulong`. This removes a checked cast under the hood.

Also plumbs through `resumeAtTopicSequenceNumber` (page) to the subscription wrapper.

I tested these changes on both topics examples successfully without any changes on the example side.


nand4011 (dismissed):
Good catch on the subscription request.

nand4011 (approved):

PR #595, opened 2024-12-04
chore: update list of retryable gRPC functions

nand4011:
Bring the list of retryable functions up to date.

Replace GetHashCode with a version that matched the equals method.

Fix misc warnings.

### DIFF
--- a/src/Momento.Sdk/Internal/Retry/DefaultRetryEligibilityStrategy.cs
+++ b/src/Momento.Sdk/Internal/Retry/DefaultRetryEligibilityStrategy.cs
@@ -1,10 +1,11 @@
 ﻿#pragma warning disable 1591
 using System;
 using System.Collections.Generic;
+using System.Linq;
 using Grpc.Core;
 using Microsoft.Extensions.Logging;
-using Microsoft.Extensions.Logging.Abstractions;
 using Momento.Protos.CacheClient;
+using Momento.Protos.CacheClient.Pubsub;
 using Momento.Sdk.Config.Retry;
 
 namespace Momento.Sdk.Internal.Retry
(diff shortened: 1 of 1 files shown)


src/Momento.Sdk/Internal/Retry/DefaultRetryEligibilityStrategy.cs
```diff
@@ -37,35 +38,59 @@ public class DefaultRetryEligibilityStrategy : IRetryEligibilityStrategy
             //StatusCode.DataLoss,
         };
 
-        private readonly HashSet<Type> _retryableRequestTypes = new HashSet<Type>
+        private readonly HashSet<Type> _retryableRequestTypes = new()
```
nand4011, line 41: Compare to the [Go SDK version](https://github.com/momentohq/client-sdk-go/blob/9ebbbe63aefbfdfee3c0fcbe574973980838b54b/internal/retry/eligibility_strategy.go#L16).

src/Momento.Sdk/Internal/Retry/DefaultRetryEligibilityStrategy.cs
```diff
@@ -100,20 +125,26 @@ public bool IsEligibleForRetry<TRequest>(Status status, TRequest request)
...
 
             var other = (DefaultRetryEligibilityStrategy)obj;
             return _retryableRequestTypes.SetEquals(other._retryableRequestTypes) &&
-                _retryableStatusCodes.SetEquals(other._retryableStatusCodes);
+                   _retryableStatusCodes.SetEquals(other._retryableStatusCodes);
         }
 
         public override int GetHashCode()
         {
-            return base.GetHashCode();
```
nand4011, line 115: We have several classes that use `base.GetHashCode()`, which violates the contract that two objects that equal each other should have the same hash code. They are mostly unimportant like this one, but we should update them when we see them.

malandis (approved):

PR #600, opened 2025-03-20
feat: add momento-local middleware for retry tests and more conversion methods to/from rpc enum and sdk error code enum

anitarua:
Work towards https://github.com/momentohq/dev-eco-issue-tracker/issues/1191 

- Adds the MomentoLocalMiddleware class and tests to validate its usage.
- Adds converter from string to RpcMethodEnum
- Adds converter from RpcMethodEnum to a string value momento-local will recognize
- Adds converter from MomentoErrorCode enum to string
- Minor changes: move a file, suppress unnecessary warnings, minor syntax improvements, add missing `WithAdditionalMiddlewares` method to `IConfiguration`
- Adds momento-local tests to PR checks in CI workflow, but filters them out from the `prod-tests` target
- Adds tests for `FixedCountRetryStrategy`


tests/Integration/Momento.Sdk.Tests/Retry/utils/MomentoLocalMiddleware.cs
```diff
@@ -0,0 +1,160 @@
...
+    /// </summary>
+    private readonly MomentoLocalMiddlewareArgs _args;
+    /// <summary>
+    /// Metrics collector class for documenting timestamps of retries for test assertions.
+    /// </summary>
+    public TestRetryMetricsCollector TestMetricsCollector { get; set; } = new TestRetryMetricsCollector();
+    /// <summary>
+    /// ID to uniquely identify api calls in a test.
+    /// </summary>
+    public string RequestId { get; set; } = Utils.NewGuidString();
```
anitarua, line 68: Like Pete pointed out on the Go PR, the middleware can be responsible for creating its own instance of a metrics collector.
I also made it create its own RequestId since each test will use a new middleware and cache client anyway.

tests/Integration/Momento.Sdk.Tests/Retry/utils/MomentoLocalMiddleware.cs
```diff
@@ -0,0 +1,161 @@
...
+using Momento.Sdk.Config.Retry;
+
+namespace Momento.Sdk.Tests.Integration.Retry;
+
+public class MomentoLocalMiddlewareArgs
+{
+    /// <summary>
+    /// Error to return from momento-local.
+    /// </summary>
+    public MomentoErrorCode? ReturnError { get; set; } = null;
```
rishtigupta, line 16: Do we want this to be of type string instead of MomentoErrorCode?
JS ref - https://github.com/momentohq/client-sdk-javascript/blob/main/packages/client-sdk-nodejs/test/momento-local-middleware.ts#L161

afair, it was such that if we expose it publicly to users, they are not tied to just testing with momentoErrorCodes.

tests/Integration/Momento.Sdk.Tests/Retry/utils/MomentoLocalMiddleware.cs
```diff
@@ -0,0 +1,161 @@
...
+    /// </summary>
+    public List<MomentoRpcMethod>? ErrorRpcList { get; set; } = null;
+    /// <summary>
+    /// The number of requests for which to return the configured error.
+    /// </summary>
+    public int? ErrorCount { get; set; } = null;
+    /// <summary>
+    /// The RPCs for which to delay the response.
+    /// </summary>
+    public List<MomentoRpcMethod>? DelayRpcList { get; set; } = null;
```
rishtigupta, line 28: same comment as above; `List<String>` instead of `List<MomentoRPCMethod>`

tests/Integration/Momento.Sdk.Tests/Retry/utils/MomentoLocalMiddleware.cs
```diff
@@ -0,0 +1,161 @@
...
+    /// </summary>
+    public int? DelayMillis { get; set; } = null;
+    /// <summary>
+    /// The number of requests for which to delay the response.
+    /// </summary>
+    public int? DelayCount { get; set; } = null;
+    /// <summary>
+    /// The RPCs for which to return an error. Currently the only stream method is TopicSubscribe.
+    /// </summary>
+    public List<MomentoRpcMethod>? StreamErrorRpcList { get; set; } = null;
```
rishtigupta, line 40: same comment as above; `List<String>` instead of `List<MomentoRPCMethod>`

tests/Integration/Momento.Sdk.Tests/Retry/utils/MomentoLocalMiddleware.cs
```diff
@@ -0,0 +1,161 @@
...
+    /// <summary>
+    /// The error to return from within an active stream.
+    /// </summary>
+    public MomentoErrorCode? StreamError { get; set; } = null;
```
rishtigupta, line 44: same comment as above; `String` instead of `MomentoErrorCode`?

tests/Integration/Momento.Sdk.Tests/Retry/utils/MomentoLocalMiddleware.cs
rishtigupta: 💪

tests/Integration/Momento.Sdk.Tests/Retry/utils/MomentoLocalMiddleware.cs
anitarua: ah that's right, good catch, will update this!

rishtigupta: sure

rishtigupta (dismissed):
:shipit:

tests/Integration/Momento.Sdk.Tests/Retry/utils/MomentoLocalMiddleware.cs
anitarua: added this [backlog ticket](https://github.com/momentohq/dev-eco-issue-tracker/issues/1202) for implementing publicly accessible version in future

anitarua: on second thought, making them strings from the start will make it consistent with the other sdks, will switch to strings for these fields and move the conversion to the tests themselves

rishtigupta: I checked Java implementation and found it to be similar to what you have implemented here: [code ref](https://github.com/momentohq/client-sdk-java/blob/60da1212bf9947474a4a40b933b659420c188b64/momento-sdk/src/intTest/java/momento/sdk/retry/utils/MomentoLocalMiddlewareArgs.java#L23).

anitarua: ah, I just saw in Pete's Go PR that it was implemented as strings so thought I should just make it consistent.
I think it's not the biggest deal either way though since it's just for tests

src/Momento.Sdk/Config/IConfiguration.cs
```diff
@@ -38,6 +38,13 @@ public interface IConfiguration
     /// <returns>Configuration object with custom middlewares provided</returns>
     public IConfiguration WithMiddlewares(IList<IMiddleware> middlewares);
 
+    /// <summary>
+    /// Add the specified middlewares to an existing instance of Configuration object in addition to already specified middlewares.
+    /// </summary>
+    /// <param name="additionalMiddlewares">The Middleware interface allows the Configuration to provide a higher-order function that wraps all requests.</param>
+    /// <returns>Configuration object with custom middlewares provided</returns>
+    public Configuration WithAdditionalMiddlewares(IList<IMiddleware> additionalMiddlewares);
```
rishtigupta, line 46: I thought other sdks use `addMiddleware` for this?
JS SDK ref: [link](https://github.com/momentohq/client-sdk-javascript/blob/main/packages/client-sdk-nodejs/src/config/configuration.ts#L112)

src/Momento.Sdk/Config/Retry/MomentoRpcMethod.cs
```diff
@@ -1,3 +1,4 @@
+#pragma warning disable 1591
```
rishtigupta, line 1: Beginner question: what does this disable?

anitarua: I found this line for disabling warnings for missing comments on all the public-facing enum cases 😅 I felt it was straightforward enough I didn't need to add a comment for every single RPC listed in the enum

src/Momento.Sdk/Config/IConfiguration.cs
anitarua: I think so, but there was already an implementation called `WithAdditionalMiddlewares` in [Configuration](https://github.com/momentohq/client-sdk-dotnet/blob/main/src/Momento.Sdk/Config/Configuration.cs#L77) that just wasn't exposed in the IConfiguration interface

rishtigupta (dismissed):
:shipit: 🚢

tests/Integration/Momento.Sdk.Tests/Retry/utils/MomentoLocalMiddleware.cs
```diff
@@ -0,0 +1,166 @@
...
+        }
+
+        var headers = callOptionsWithHeaders.Headers!;
+
+        if (headers.GetValue("request-id") == null)
+        {
+            _logger.LogDebug($"MomentoLocalMiddleware: requestId={RequestId}");
+            headers.Add("request-id", RequestId);
+        }
+        if (_args.ReturnError != null && headers.GetValue("return-error") == null)
```
nand4011, line 94: Why do we need to check if the header is already there? I don't recall doing that in the other implementations.

nand4011: This is a style thing, but with an extension method you could get rid of the header null check:
```
public static class MetadataExtensions
{
    public static void TryAddHeader(this Metadata metadata, string key, string value)
    {
        if (metadata.GetValue(key) == null)
        {
            metadata.Add(key, value);
        }
    }
}
```
```
if (_args.ReturnError is not null)
 {
    _logger.LogDebug("MomentoLocalMiddleware: return-error={returnError}", _args.ReturnError);
    headers.TryAddHeader("return-error", _args.ReturnError);
}
```

nand4011 (commented):
Couple small things, but LGTM otherwise.

src/Momento.Sdk/Config/Retry/MomentoRpcMethod.cs
```diff
@@ -110,7 +110,122 @@ public static string ToStringValue(this MomentoRpcMethod rpcMethod)
...
+        };
+    }
+
+    /// <summary>
+    /// Converts the rpc method to a string value recognized by momento-local.
+    /// </summary>
+    /// <param name="rpcMethod">to convert to a string</param>
+    /// <returns></returns>
+    /// <exception cref="ArgumentOutOfRangeException">if given an unknown rpc method</exception>
+    public static string ToMomentoLocalMetadataString(this MomentoRpcMethod rpcMethod)
```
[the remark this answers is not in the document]
anitarua: good point, I had put it there because JS had it in the public code, but that's bc JS also has a publicly accessible momento-local middleware whereas that's not a thing here yet. Will move this to test folder for now

tests/Integration/Momento.Sdk.Tests/Retry/utils/MomentoLocalMiddleware.cs
```diff
@@ -0,0 +1,166 @@
...
+        {
+            _logger.LogDebug($"MomentoLocalMiddleware: return-error={_args.ReturnError}");
+            // headers.Add("return-error", MomentoErrorCodeMetadataConverter.ToStringValue((MomentoErrorCode)_args.ReturnError));
```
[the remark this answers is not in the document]
anitarua: whoops, meant to clean this up after making the middleware args change to accept strings instead

tests/Integration/Momento.Sdk.Tests/Retry/utils/MomentoLocalMiddleware.cs
anitarua: I ran into some errors when I didn't have the null checks and requests were getting retried. I assumed "Add" was always appending a new key-pair value into the Metadata object rather than overwriting a key if it existed.

nand4011 (approved):

PR #601, opened 2025-03-24
feat: add FixedTimeoutRetryStrategy and tests

anitarua:
Closes https://github.com/momentohq/dev-eco-issue-tracker/issues/1195


src/Momento.Sdk/Internal/Middleware/RetryMiddleware.cs
```diff
@@ -60,6 +67,24 @@ Func<TRequest, CallOptions, Task<MiddlewareResponseState<TResponse>>> continuati
...
+                        
+                        // Check if new deadline would exceed overall deadline. Use overall if so.
+                        if (retryDeadline > _overallDeadline)
+                        {
+                            callOptions = callOptions.WithDeadline(_overallDeadline);
+                        } else
+                        {
+                            callOptions = callOptions.WithDeadline(retryDeadline);
+                        }
+                    }
```
anitarua, line 87: I think the JS and Java implementations may do something slightly differently when determining whether to set the next deadline, and we should probably decide which approach is best and make sure they all behave similarly.
- JS: on each retry attempt, check if current time has passed overall deadline yet ([reference](https://github.com/momentohq/client-sdk-javascript/blob/main/packages/client-sdk-nodejs/src/internal/grpc/retry-interceptor.ts#L100))
- Java: check if deadline is expired on every retry attempt ([reference](https://github.com/momentohq/client-sdk-java/blob/main/momento-sdk/src/main/java/momento/sdk/RetryClientInterceptor.java#L119)) and also check if next retry deadline would exceed overall deadline, timeout immediately ([reference](https://github.com/momentohq/client-sdk-java/blob/main/momento-sdk/src/main/java/momento/sdk/RetryClientInterceptor.java#L163))

I'm not certain but it feels like we shouldn't have to manually check if the deadline has expired in our retry interceptors, that grpc should handle that for us. 
I think the more correct thing would be to determine when to set a request's deadline as the overall deadline rather than the retry deadline, then let grpc handle the timeout when it occurs.

src/Momento.Sdk/Config/Retry/IRetryStrategy.cs
```diff
@@ -1,5 +1,5 @@
+using System;
```
rishtigupta, line 1: Why do we need this import?

anitarua: ah it's not needed, probably leftover from some earlier twiddling/experimenting, will remove

src/Momento.Sdk/Internal/Middleware/RetryMiddleware.cs
anitarua: From conversation: no need to manually timeout a retry attempt, except for Java retry interceptor which doesn't know about the deadline

anitarua:
Closing in favor of https://github.com/momentohq/client-sdk-dotnet/pull/602

Will make additional tickets according to this [alignment discussion](https://github.com/momentohq/client-sdk-dotnet/pull/601#discussion_r2011062001)

PR #603, opened 2025-03-27
feat: add ExponentialBackoffRetryStrategy and fix test middleware timestamp bug

anitarua:
Work towards https://github.com/momentohq/dev-eco-issue-tracker/issues/1195

Adds ExponentialBackoffRetryStrategy and tests based on the JS, Java, and Go implementations.

Also found a bug in how the MomentoLocalMiddleware was recording timestamps (it was not, it was only recording the number 1 🤦)

### DIFF
--- a/src/Momento.Sdk/Config/Retry/ExponentialBackoffRetryStrategy.cs
+++ b/src/Momento.Sdk/Config/Retry/ExponentialBackoffRetryStrategy.cs
@@ -1,0 +1,129 @@
+using System;
+using Grpc.Core;
+using Microsoft.Extensions.Logging;
+using Momento.Sdk.Internal.Retry;
+
+namespace Momento.Sdk.Config.Retry;
+
+/// <summary>
+/// TODO
+/// </summary>
+public class ExponentialBackoffRetryStrategy : IRetryStrategy
+{
+    /// <summary>
+    /// TODO
+    /// </summary>
+    public static readonly double DEFAULT_INITIAL_DELAY_MS = 0.5;
+    /// <summary>
+    /// TODO
+    /// </summary>
+    public static readonly double DEFAULT_GROWTH_FACTOR = 2;
+    /// <summary>
+    /// TODO
+    /// </summary>
+
+    public static readonly double DEFAULT_MAX_BACKOFF_MS = 1000;
+    private ILoggerFactory _loggerFactory;
+    private ILogger _logger;
+    private readonly IRetryEligibilityStrategy _eligibilityStrategy;
+    
+    private readonly double _initialDelayMillis;
+    private readonly double _growthFactor;
+    private readonly double _maxBackoffMillis;
+
+    /// <summary>
+    /// TODO
+    /// </summary>
+    /// <param name="loggerFactory"></param>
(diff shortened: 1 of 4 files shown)


src/Momento.Sdk/Config/Retry/ExponentialBackoffRetryStrategy.cs
```diff
@@ -0,0 +1,135 @@
...
+    private readonly Random _random = new Random();
+
+    /// <summary>
+    /// Constructor for ExponentialBackoffRetryStrategy
+    /// </summary>
+    /// <param name="loggerFactory"></param>
+    /// <param name="eligibilityStrategy"></param>
+    /// <param name="initialDelayMillis"></param>
+    /// <param name="maxBackoffMillis"></param>
+    public ExponentialBackoffRetryStrategy(ILoggerFactory loggerFactory, IRetryEligibilityStrategy? eligibilityStrategy = null, double? initialDelayMillis = null, double? maxBackoffMillis = null)
```
malandis, line 47: Could make sense to use `TimeSpan` to type anything related to time in the API

src/Momento.Sdk/Config/Retry/ExponentialBackoffRetryStrategy.cs
```diff
@@ -0,0 +1,135 @@
...
+    /// </summary>
+    public static readonly double DEFAULT_MAX_BACKOFF_MS = 1000;
+
+    private ILoggerFactory _loggerFactory;
+    private ILogger _logger;
+    private readonly IRetryEligibilityStrategy _eligibilityStrategy;
+    
+    private readonly double _initialDelayMillis;
+    private readonly double _growthFactor;
+    private readonly double _maxBackoffMillis;
```
malandis, line 37: Could make these `TimeSpan`s too

src/Momento.Sdk/Config/Retry/ExponentialBackoffRetryStrategy.cs
```diff
@@ -0,0 +1,135 @@
...
+        if (!_eligibilityStrategy.IsEligibleForRetry(grpcStatus, grpcRequest))
+        {
+            return null;
+        }
+
+        double baseDelay = computeBaseDelay(attemptNumber);
+        double previousBaseDelay = computePreviousBaseDelay(baseDelay);
+        double maxDelay = previousBaseDelay * 3;
+        double jitteredDelay = randomInRange(baseDelay, maxDelay);
+        int jitteredDelayMs = Convert.ToInt32(jitteredDelay);
```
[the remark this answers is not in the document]
anitarua: `Convert.ToInt32` already handles rounding based on the description I see:

> int Convert.ToInt32(double value) (+ 18 overloads)
Converts the value of the specified double-precision floating-point number to an equivalent 32-bit signed integer.
Returns:
  value, rounded to the nearest 32-bit signed integer. If value is halfway between two whole numbers, the even number is returned; that is, 4.5 is converted to 4, and 5.5 is converted to 6.

malandis (approved):

PR #604, opened 2025-03-27
feat: add TopicClient middleware and topics tests against momento-local

anitarua:
Closes https://github.com/momentohq/dev-eco-issue-tracker/issues/1189

Adds middleware to the TopicClient config and internals (stream connections provide only an option for adding headers).
Adds topic client retry tests for exercising subscription reconnect logic ([JS reference](https://github.com/momentohq/client-sdk-javascript/blob/main/packages/client-sdk-nodejs/test/integration/retry/topic-client-retry.test.ts)).

Would really appreciate feedback on a couple of points:

1. Topics retry tests: subscriptions don't follow the JS/Java callback approach so there's no usage of semaphores or an `OnConnectionLost()` callback here. Not sure if there's a better way to track subscription events for tests.
2. Not sure if the outlined approach is the most idiomatic. Maybe an IStreamMiddleware interface could split out the stream headers part? But if the current design seems good enough, we can also proceed with what's here.

### DIFF
--- a/src/Momento.Sdk/Config/Middleware/IMiddleware.cs
+++ b/src/Momento.Sdk/Config/Middleware/IMiddleware.cs
@@ -1,4 +1,6 @@
 using System;
+using System.Collections;
+using System.Collections.Generic;
 using System.Threading.Tasks;
 using Grpc.Core;
 using Grpc.Core.Interceptors;
@@ -37,6 +39,11 @@ public record struct MiddlewareResponseState<TResponse>(
 public interface IMiddleware
 {
     /// <summary>
+    /// Add headers to gRPC stream requests since middlewares are incompatible with gRPC streams.
+    /// </summary>
+    public IList<Tuple<string, string>> AddStreamRequestHeaders();
+
+    /// <summary>
     /// Called as a wrapper around each request; can be used to time the request and collect metrics etc.
     /// </summary>
     /// <typeparam name="TRequest"></typeparam>
@@ -51,5 +58,4 @@ public interface IMiddleware
         CallOptions callOptions,
         Func<TRequest, CallOptions, Task<MiddlewareResponseState<TResponse>>> continuation
     ) where TRequest : class where TResponse : class;
-
 }
--- a/src/Momento.Sdk/Config/TopicConfiguration.cs
+++ b/src/Momento.Sdk/Config/TopicConfiguration.cs
@@ -1,5 +1,8 @@
 using System;
+using System.Collections.Generic;
+using System.Linq;
 using Microsoft.Extensions.Logging;
+using Momento.Sdk.Config.Middleware;
 using Momento.Sdk.Config.Transport;
 
 namespace Momento.Sdk.Config;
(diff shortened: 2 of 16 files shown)


rishtigupta (approved):
:shipit:

tests/Integration/Momento.Sdk.Tests/Retry/FixedCountRetryStrategyTest.cs
```diff
@@ -39,30 +40,40 @@ public void FixedCountRetryStrategy_IneligibleRpc()
...
+    public async Task FixedCountRetryStrategy_EligibleRpc_FullOutage() 
     {
         var maxAttempts = 3;
         var middlewareArgs = new MomentoLocalMiddlewareArgs {
             ReturnError = MomentoErrorCode.SERVER_UNAVAILABLE.ToStringValue(),
             ErrorRpcList = new List<string> { MomentoRpcMethod.Get.ToMomentoLocalMetadataString() },
         };
         var testProps = new MomentoLocalCacheAndCacheClient(_authProvider, _loggerFactory, _cacheConfig, middlewareArgs, new FixedCountRetryStrategy(_loggerFactory, maxAttempts));
-        testProps.CacheClient.GetAsync(testProps.CacheName, "key").Wait();
+        var result = await testProps.CacheClient.GetAsync(testProps.CacheName, "key");
```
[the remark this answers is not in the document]
anitarua: I think it's related to whether the function is marked as async or not. `Wait()` could be used instead of `await`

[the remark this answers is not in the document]
rishtigupta: That's fine. I just wanted to confirm the difference. 
Thanks!

src/Momento.Sdk/Config/Middleware/IMiddleware.cs
```diff
@@ -36,6 +38,11 @@ Func<Metadata> GetTrailers
 /// </summary>
 public interface IMiddleware
 {
+    /// <summary>
+    /// Add headers to gRPC stream requests since middlewares are incompatible with gRPC streams.
+    /// </summary>
+    public IList<Tuple<string, string>> AddStreamRequestHeaders();
+
```
malandis, line 45: Why do we need this? All of the implementations are trivial, ie return an empty list.

src/Momento.Sdk/Config/TopicConfiguration.cs
```diff
@@ -12,33 +15,61 @@ public class TopicConfiguration : ITopicConfiguration
 
     /// <inheritdoc />
     public ITopicTransportStrategy TransportStrategy { get; }
+    /// <inheritdoc />
+    public IList<IMiddleware> Middlewares { get; }
+    /// <inheritdoc />
+    public IList<Tuple<string, string>> SubscribeHeaders { get; }
```
malandis, line 21: Why do we need this? Is there ever a case where this is initialized?

src/Momento.Sdk/Internal/TopicGrpcManager.cs
```diff
@@ -51,16 +50,33 @@ public async Task<_Empty> publish(_PublishRequest request, CallOptions callOptio
...
         {
-            callHeaders.Add(header.Item1, header.Item2);
+            if (callHeaders.Get(header.Item1) == null)
+            {
+                callHeaders.Add(header.Item1, header.Item2);
+            }
+        }
+
+        // We also add any stream headers configured on the middlewares.
+        foreach (var middleware in _middlewares)
```
malandis, line 71: None of the middlewares have headers. Can we get rid of this?

anitarua:
I think the ^ questions all stem from my trying to make the `IMiddleware` interface work for both unary and stream connections, but this [comment](https://github.com/momentohq/client-sdk-dotnet/blob/main/src/Momento.Sdk/Internal/TopicGrpcManager.cs#L53-L54) states that the current middleware setup is incompatible with streaming calls so we must manually add headers.

I'm gonna put this back in draft to see if I could extract the stream headers bit into another interface and have the topic client accept stream-compatible middleware to allow us to write momento-local tests against topics

anitarua:
Alright, I think I found a better implementation, closing this PR in favor of http://github.com/momentohq/client-sdk-dotnet/pull/609

@rishtigupta I can help rebase the PR that was based on this one

rishtigupta:
> Alright, I think I found a better implementation, closing this PR in favor of [http://github.com/momentohq/client-sdk-dotnet/pull/609](https://github.com/momentohq/client-sdk-dotnet/pull/609)
> 
> @rishtigupta I can help rebase the PR that was based on this one

Thanks for the heads up. I wanted your implementation just for the tests. So, once we decide on the topic middleware, it should be easy for me to rebase my pr.

PR #602, opened 2025-03-27
feat: fixed timeout retry strategy and tests

anitarua:
Work towards https://github.com/momentohq/dev-eco-issue-tracker/issues/1195

Adds fixed timeout retry strategy (used [JS](https://github.com/momentohq/client-sdk-javascript/blob/main/packages/client-sdk-nodejs/src/config/retry/fixed-timeout-retry-strategy.ts) and [Java](https://github.com/momentohq/client-sdk-java/blob/main/momento-sdk/src/main/java/momento/sdk/retry/FixedTimeoutRetryStrategy.java) implementations for reference).
Adds momento-local tests exercising the fixed timeout strategy.

PR #606, opened 2025-04-07
fix: fixed timeout retry strategy was stopping retries too early when retry deadline passed

anitarua:


src/Momento.Sdk/Internal/Middleware/RetryMiddleware.cs
```diff
@@ -62,8 +61,11 @@ Func<TRequest, CallOptions, Task<MiddlewareResponseState<TResponse>>> continuati
                     var status = nextState.GetStatus();
                     _logger.LogDebug($"Request failed with status {status.StatusCode}, checking to see if we should retry; attempt Number: {attemptNumber}");
                     _logger.LogTrace($"Failed request status: {status}");
-                    retryAfterMillis = _retryStrategy.DetermineWhenToRetryRequest(nextState.GetStatus(), request, attemptNumber);
-                    callOptions = CalculateRetryDeadline(callOptions, _overallDeadline);
+                    retryAfterMillis = _retryStrategy.DetermineWhenToRetryRequest(nextState.GetStatus(), request, attemptNumber, _overallDeadline);
+                    if (_retryStrategy is FixedTimeoutRetryStrategy strategy)
+                    {
+                        callOptions = strategy.CalculateRetryDeadline(callOptions, _overallDeadline);
+                    }
```
malandis, line 68: This code shouldn't be aware of the concrete retry strategy implementations. If we need additional refinement then we should have an interface method on the retry strategy `CalculateRetryDeadline` that has a default implementation for all strategies and the specific one for `FixedTimeoutRetryStrategy`.

src/Momento.Sdk/Config/Retry/IRetryStrategy.cs
```diff
@@ -7,18 +8,13 @@ namespace Momento.Sdk.Config.Retry;
...
-    public double? GetResponseDataReceivedTimeoutMillis();
+  /// <summary>
+  /// Calculates whether or not to retry a request based on the type of request and number of attempts.
+  /// </summary>
+  /// <param name="grpcStatus"></param>
+  /// <param name="grpcRequest"></param>
+  /// <param name="attemptNumber"></param>
+  /// <param name="overallDeadline"></param>
+  /// <returns>Returns number of milliseconds after which the request should be retried, or <see langword="null"/> if the request should not be retried.</returns>
+  public int? DetermineWhenToRetryRequest<TRequest>(Status grpcStatus, TRequest grpcRequest, int attemptNumber, DateTime overallDeadline) where TRequest : class;
```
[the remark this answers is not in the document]
anitarua: This was [added last week](https://github.com/momentohq/client-sdk-dotnet/commit/ddd789abe0dc1003815ed1489852ce1351afe439) and there hasn't been a release of it yet, does it still count as a breaking change?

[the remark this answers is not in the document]
anitarua: Thanks for catching this! I made the new parameter optional, I think that should make it backwards compatible

src/Momento.Sdk/Internal/Middleware/RetryMiddleware.cs
anitarua: good call, I made the change, `CalculateRetryDeadline` now exists on `IRetryStrategy` and each strategy has its own implementation

src/Momento.Sdk/Config/Retry/IRetryStrategy.cs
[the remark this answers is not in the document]
malandis: Doesn't that make the method ambiguous with the other one when no `overallDeadline` is provided?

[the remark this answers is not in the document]
anitarua: There's only the one `DetermineWhenToRetryRequest` method right now. 
I could make another with the extra parameter, but I updated the retry middleware to call only the version with the deadline and all the retry strategies would end up needing to implement a version of both

src/Momento.Sdk/Config/Retry/IRetryStrategy.cs
```diff
@@ -15,10 +16,29 @@ public interface IRetryStrategy
     /// <param name="attemptNumber"></param>
     /// <returns>Returns number of milliseconds after which the request should be retried, or <see langword="null"/> if the request should not be retried.</returns>
     public int? DetermineWhenToRetryRequest<TRequest>(Status grpcStatus, TRequest grpcRequest, int attemptNumber) where TRequest : class;
+}
+
+/// <summary>
+/// Defines a contract for how and when to retry a request, including the ability to set a deadline for the retry.
+/// This is useful for strategies like the FixedTimeoutRetryStrategy that overwrites the retry deadline.
+/// </summary>
+public interface IDeadlineAwareRetryStrategy : IRetryStrategy
```
anitarua, line 25: made a new interface, so the IRetryStrategy should match what's already been released. 
new strategy includes deadline-aware DetermineWhenToRetryRequest as well as new method CalculateRetryDeadline, replacing GetResponseDataReceivedTimeoutMillis (which is currently on IRetryStrategy in main but unreleased).

src/Momento.Sdk/Internal/Middleware/RetryMiddleware.cs
```diff
@@ -62,8 +63,18 @@ Func<TRequest, CallOptions, Task<MiddlewareResponseState<TResponse>>> continuati
                     var status = nextState.GetStatus();
                     _logger.LogDebug($"Request failed with status {status.StatusCode}, checking to see if we should retry; attempt Number: {attemptNumber}");
                     _logger.LogTrace($"Failed request status: {status}");
-                    retryAfterMillis = _retryStrategy.DetermineWhenToRetryRequest(nextState.GetStatus(), request, attemptNumber);
-                    callOptions = CalculateRetryDeadline(callOptions, _overallDeadline);
+
+                    if (_retryStrategy is IDeadlineAwareRetryStrategy strategy)
```
anitarua, line 67: asserting which interface we're using here

src/Momento.Sdk/Config/Retry/FixedCountRetryStrategy.cs
```diff
@@ -1,6 +1,7 @@
 using Grpc.Core;
 using Microsoft.Extensions.Logging;
 using Momento.Sdk.Internal.Retry;
+using System;
```
rishtigupta, line 4: Can be removed?

anitarua: yes, will clean that up

src/Momento.Sdk/Config/Retry/IRetryStrategy.cs
```diff
@@ -1,4 +1,5 @@
 using Grpc.Core;
+using System;
```
rishtigupta, line 2: this one too.

src/Momento.Sdk/Config/Retry/FixedTimeoutRetryStrategy.cs
```diff
@@ -87,8 +104,13 @@ public override int GetHashCode()
     }
 
     /// <inheritdoc/>
-    public double? GetResponseDataReceivedTimeoutMillis()
+    public int? DetermineWhenToRetryRequest<TRequest>(Status grpcStatus, TRequest grpcRequest, int attemptNumber) where TRequest : class
```
malandis, line 107: It's a weird corner case but should comment this won't be invoked by the retry middleware.

src/Momento.Sdk/Internal/Middleware/RetryMiddleware.cs
```diff
@@ -29,6 +29,7 @@ Func<TRequest, CallOptions, Task<MiddlewareResponseState<TResponse>>> continuati
             // The first time we enter WrapRequest, we capture the overall deadline for the request
             // in case we're using the FixedTimeoutRetryStrategy
             DateTime _overallDeadline = callOptions.Deadline ?? Utils.CalculateDeadline(_clientTimeout);
+            Console.WriteLine($"Overall deadline: {_overallDeadline}");
```
rishtigupta, line 32: Is the print statement intentional?

src/Momento.Sdk/Config/Retry/IRetryStrategy.cs
anitarua: oh this one caused a complaint when I tried removing it, was needed for DateTime

src/Momento.Sdk/Internal/Middleware/RetryMiddleware.cs
anitarua: lol no, will thanks for catching that

rishtigupta: probably good to use logger if so.

src/Momento.Sdk/Internal/Middleware/RetryMiddleware.cs
```diff
@@ -62,8 +63,18 @@ Func<TRequest, CallOptions, Task<MiddlewareResponseState<TResponse>>> continuati
...
                     _logger.LogDebug($"Request failed with status {status.StatusCode}, checking to see if we should retry; attempt Number: {attemptNumber}");
                     _logger.LogTrace($"Failed request status: {status}");
-                    retryAfterMillis = _retryStrategy.DetermineWhenToRetryRequest(nextState.GetStatus(), request, attemptNumber);
-                    callOptions = CalculateRetryDeadline(callOptions, _overallDeadline);
+
+                    if (_retryStrategy is IDeadlineAwareRetryStrategy strategy)
+                    {
+                        retryAfterMillis = strategy.DetermineWhenToRetryRequest(status, request, attemptNumber, _overallDeadline);
+                        callOptions = strategy.CalculateRetryDeadline(callOptions, _overallDeadline);
+                        Console.WriteLine($"Retrying request with retry deadline {callOptions.Deadline}");
```
rishtigupta, line 71: remove print statements!

malandis (commented):
Great last change. I just recommend commenting on the fixed timeout retyr strategy to indicate the version of determine when to retry without the overall deadline won't be invoked.

src/Momento.Sdk/Config/Retry/IRetryStrategy.cs
rishtigupta: `IDeadlineAwareRetryStrategy : IRetryStrategy` what does this syntax in dotnet mean?
is it extending the previous interface?

src/Momento.Sdk/Config/Retry/FixedTimeoutRetryStrategy.cs
anitarua: oh, I wrote it such that the version with deadline will do its deadline-specific check, but otherwise it would call the other version to do the usual checks (using eligibility strategy and such), so it's kind of still used by fixed timeout retry strategy?

src/Momento.Sdk/Config/Retry/IRetryStrategy.cs
anitarua: yes, it's implementing the IRetryStrategy so that it can still be accepted as retry strategy in places where IRetryStrategy is used

src/Momento.Sdk/Config/Retry/FixedTimeoutRetryStrategy.cs
malandis: I mean since we're using interface inheritance, the fixed timeout retry strategy has to define a version of `DetermineWhenToRetry` that will never be used.

src/Momento.Sdk/Config/Retry/ExponentialBackoffRetryStrategy.cs
```diff
@@ -13,7 +13,7 @@ namespace Momento.Sdk.Config.Retry;
 /// the current backoff and 3 times the previous backoff, with the 
 /// current backoff capped at `maxBackoffMillis`
 /// </summary>
-public class ExponentialBackoffRetryStrategy : IRetryStrategy
+public class ExponentialBackoffRetryStrategy : IBasicRetryStrategy
```
rishtigupta, line 16: changing the name of the interface `IRetryStrategy` to `IBasicretryStrategy` would be a breaking change too? 
Ideally, we would like to create a base class, and then let the existing IretryStrategy implement that base class?

anitarua: this was to minimize breaking changes to the Configuration objects which accept `IRetryStrategy` 😅

rishtigupta: Can we confirm with the team on this since I thought the whole idea was to not make any breaking changes? 😢

nand4011: If IDeadlineAwareRetryStrategy is an IRetryStrategy, why do we need a separate IBasicRetryStrategy?

src/Momento.Sdk/Config/Retry/FixedTimeoutRetryStrategy.cs
```diff
@@ -43,9 +43,20 @@ public FixedTimeoutRetryStrategy(ILoggerFactory loggerFactory, IRetryEligibility
...
     /// <inheritdoc/>
-    public int? DetermineWhenToRetryRequest<TRequest>(Status grpcStatus, TRequest grpcRequest, int attemptNumber) where TRequest : class
+    public int? DetermineWhenToRetryRequest<TRequest>(Status grpcStatus, TRequest grpcRequest, int attemptNumber, DateTime overallDeadline) where TRequest : class
     {
         _logger.LogDebug($"Determining whether request is eligible for retry; status code: {grpcStatus.StatusCode}, request type: {grpcRequest.GetType()}, attemptNumber: {attemptNumber}");
+
+        // If a retry attempt's timeout has passed but the client's overall timeout has not yet passed,
+        // we should reset the deadline and retry.
+        // Note: dotnet appears to return Cancelled instead of DeadlineExceeded when the deadline passes.
+        if (attemptNumber > 1 && overallDeadline != null && overallDeadline > DateTime.UtcNow
```
nand4011, line 53: If we want to stick to the overall deadline, this shouldn't make a new attempt if the overall deadline is closer than a retry delay interval.

anitarua:
@nand4011 @malandis 

To recap the state of this PR:

- The last sdk release contains `IRetryStrategy` with only `DetermineWhenToRetryRequest()` defined on it. We want to avoid breaking changes to this interface, which means avoiding adding anything new.
- I had already added something new (`GetResponseDataReceivedTimeoutMillis()`) in a [previous PR](https://github.com/momentohq/client-sdk-dotnet/pull/602) but this has not been released yet, so we still have the opportunity to fix it by removing `GetResponseDataReceivedTimeoutMillis()` from `IRetryStrategy` and its implementations.
- This PR creates a new interface `IDeadlineAwareRetryStrategy : IRetryStrategy` instead for the `FixedTimeoutRetryStrategy`. The new interface defines a `CalculateRetryDeadline()` method and uses a `DetermineWhenToRetryRequest()` that accepts `overallDeadline` as a parameter. 
- The `RetryMiddleware` calls the appropriate `DetermineWhenToRetryRequest()` based on what type of retry strategy interface it receives.
- Finally, this PR adds a new test and fixes a couple of previously written fixed timeout retry strategy tests.

End result should be cleaner implementation of `FixedTimeoutRetryStrategy` and bug fix for incorrect behavior (deadline was not properly overwritten with each retry when using this strategy).

PR #607, opened 2025-04-07
chore: add lint and format Makefile targets and run linter in CI

anitarua:
Found [`dotnet format`](https://learn.microsoft.com/en-us/dotnet/core/tools/dotnet-format) to use as a linter/formatter!


Makefile
```diff
@@ -59,6 +59,18 @@ all: precommit
 build: ${BUILD_TARGETS}
 
 
+## Lint the project
+lint:
+	@echo "Linting the project..."
```
malandis, line 64: This is more like `format-check`

Makefile
```diff
@@ -59,6 +59,18 @@ all: precommit
...
+	@dotnet format --verify-no-changes
+
+
+## Apply formatting to the project
+format:
+	@echo "Applying formatting to the project..."
+	@dotnet format
```
malandis, line 71: Is there a reference for configuring the format rules/settings?

malandis: I mean what rules does `dotnet format` apply? Can you externalize those to a file? Can we hook up the IDE to delegate the formatting to this?

anitarua: oh hm, I'm not sure what the defaults are, tried following the link and doing a quick search but didn't find anything specific:

> Preferences will be read from an .editorconfig file, if present, otherwise a default set of preferences will be used. For more information, see the [EditorConfig documentation](https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/configuration-files#editorconfig).

anitarua: For IDE auto-formatting, I found this setting that can be added to vscode's settings.json:
```
{
    "editor.formatOnSave": true,
}
```

Makefile
malandis: cc @anitarua

malandis (approved):

PR #609, opened 2025-04-10
feat: add ITopicMiddleware and implement it for momento-local topics tests

anitarua:
Closes https://github.com/momentohq/dev-eco-issue-tracker/issues/1189

Adds a new ITopicMiddleware interface to allow TopicClient to accept middlewares that can be applied to both publish and subscribe. Implemented it for MomentoLocalMiddleware and wrote new momento-local tests for topics.

Because of this [comment](https://github.com/momentohq/client-sdk-dotnet/blob/main/src/Momento.Sdk/Internal/TopicGrpcManager.cs#L53-L54) stating that the current middleware setup is incompatible with streaming calls, I made this middleware more limited in what it can change. Namely, it can really only add headers before a request is sent out. I might be limiting the capabilities here too much, let me know what y'all think!

In addition to headers, the ITopicMiddleware also allows for taking action when a stream is established or disconnected. All other topics events are accessible through an iterator-style subscription stream, but these two events were not, but was helpful for the momento-local tests, and could be helpful for other users as well, but not as important as the main topics events provided in the regular subscription iterator.

### DIFF
--- a/src/Momento.Sdk/Config/Middleware/ITopicMiddleware.cs
+++ b/src/Momento.Sdk/Config/Middleware/ITopicMiddleware.cs
@@ -4,24 +4,24 @@ using System.Collections.Generic;
 namespace Momento.Sdk.Config.Middleware;
 
 /// <summary>
-/// TODO
+/// Interface for TopicClient middleware.
 /// </summary>
 public interface ITopicMiddleware
 {
     /// <summary>
-    /// TODO
+    /// Add arbitrary headers to each request.
     /// </summary>
     /// <returns></returns>
     public IList<Tuple<string, string>> WithHeaders();
 
     /// <summary>
-    /// TODO
+    /// Take action when a subscription stream is disconnected.
     /// </summary>
     /// <returns></returns>
     public void OnStreamDisconnected();
 
     /// <summary>
-    /// TODO
+    /// Take action when a subscription stream is (re)established.
     /// </summary>
     /// <returns></returns>
     public void onStreamEstablished();
--- a/src/Momento.Sdk/Config/Middleware/ITopicMiddleware.cs
+++ b/src/Momento.Sdk/Config/Middleware/ITopicMiddleware.cs
@@ -9,10 +9,10 @@ namespace Momento.Sdk.Config.Middleware;
 public interface ITopicMiddleware
 {
     /// <summary>
-    /// Add arbitrary headers to each request.
+    /// Arbitrary headers to each request.
     /// </summary>
     /// <returns></returns>
-    public IList<Tuple<string, string>> WithHeaders();
+    public IEnumerable<KeyValuePair<string, string>> Headers { get; }
 
     /// <summary>
     /// Take action when a subscription stream is disconnected.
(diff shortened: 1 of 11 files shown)


src/Momento.Sdk/Config/Middleware/ITopicMiddleware.cs
```diff
@@ -0,0 +1,28 @@
+using System;
+using System.Collections.Generic;
+
+namespace Momento.Sdk.Config.Middleware;
+
+/// <summary>
+/// Interface for TopicClient middleware.
+/// </summary>
+public interface ITopicMiddleware
```
rishtigupta, line 9: Should we call it IStreamMiddleware? In future, we could use this for batch requests?

anitarua: hmm, I'm not sure. 

- we're currently exposing get/set batch calls as unary requests through other SDKs
- this middleware has some things I think are topics-specific (OnStreamDisconnect and OnStreamEstablished), not sure if that would carry over to batch calls as well

rishtigupta: I see. Thanks for confirming!

rishtigupta (approved):
LGTM! Just waiting on a final review from @malandis

src/Momento.Sdk/Config/Middleware/ITopicMiddleware.cs
```diff
@@ -0,0 +1,28 @@
...
+{
+    /// <summary>
+    /// Arbitrary headers to each request.
+    /// </summary>
+    /// <returns></returns>
+    public IEnumerable<KeyValuePair<string, string>> Headers { get; }
```
nand4011, line 15: If we're going to make a streaming middleware, my first instinct would be to have it have a wrap request function that operates on a `AsyncServerStreamingCall<>` and have a StreamingHeaderMiddleware instead of making the middleware interface aware of headers. Otherwise, it feels like we're half retrofitting callbacks onto topics. I'm not opposed to that, but I think it's different than a middleware, which is supposed to be able to see and modify an entire request/response lifecycle.
I wish we had the batch calls implemented here so we could see how they could be modified to use a streaming middleware.
I think we should either:
1. Add a general streaming middleware interface that fully wraps a streaming request and could be used for the batch calls and topic publish, or
2. Add specific topic callbacks and don't call them TopicMiddleware.

@malandis do you have an opinion on this?

anitarua: good points! I think separating the topics-specific callbacks and naming them something not middleware makes sense.

As for the streaming middleware, I think the `wrapRequest` function approach makes sense, but I'll need to do some digging to verify it'll work (based on my PR description from when I last worked on this).

> Because of this [comment](https://github.com/momentohq/client-sdk-dotnet/blob/main/src/Momento.Sdk/Internal/TopicGrpcManager.cs#L53-L54) stating that the current middleware setup is incompatible with streaming calls, I made this middleware more limited in what it can change. Namely, it can really only add headers before a request is sent out. I might be limiting the capabilities here too much, let me know what y'all think!

anitarua:
Closing this PR in favor of https://github.com/momentohq/client-sdk-dotnet/pull/611 which makes only internal or test-only changes

PR #610, opened 2025-04-15
feat: add configurations for topics with separate unary and stream channels

rishtigupta:
## PR Description:
This PR introduces support for independently configuring the number of unary and streaming gRPC channels for the Topics client. It includes the following key changes:

- Added configuration support for specifying distinct values for:
    - `numStreamGrpcChannels` 
    - `numUnaryGrpcChannels` 
- Refactored `TopicGrpcManager` to create and manage separate pools of stubs for unary and streaming operations. 

## Issue
https://github.com/momentohq/dev-eco-issue-tracker/issues/1156


anitarua:
I think this PR is not needed actually. Realized we use `SocketsHttpHandler.EnableMultipleHttp2Connections = true` ([reference](https://learn.microsoft.com/en-us/aspnet/core/grpc/performance?view=aspnetcore-6.0#connection-concurrency)) which allows the .net client to create new underlying connections to support more grpc requests if needed. Using this setting also aligns with the grpc .net defaults as [Michael pointed out here](https://github.com/momentohq/client-sdk-dotnet/pull/539/files#r1506647596).

Also did some testing to confirm that the current state of main already can support >100 concurrent streams, so will close this PR and ticket as it's not needed.

